PR #23, opened 2021-08-26
Chore/carrier apps accounts

kitajchuk:
My take here is that when we have Account and Application selectors together we want the options in the Application selector to be based on the value of the Accounts selector—since Applications are scoped to Accounts.

This PR:

- Adds Account selector to Add/Edit Carrier view
- Binds the options for the Application selector to be based on the value of the Account selector for Carrier form
- Binds the options for the Application selector to be based on the value of the Account selector for Phone Numbers form
- Fixes bug when trying to add an Application under a Service Provider with no Accounts created yet (cherry-picked)
- Adds ability to delete service providers so long as you have more than one

### DIFF
--- a/src/components/forms/ApplicationForm.js
+++ b/src/components/forms/ApplicationForm.js
@@ -120,7 +120,7 @@ const ApplicationForm = props => {
           applicationsPromise,
         ]);
 
-        const accounts     = promiseAllValues[0].data;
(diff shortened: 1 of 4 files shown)


kitajchuk:
@davehorton I've finalized the frontend changes for deleting service providers after pulling your fix to the API server. You can now delete service providers so long as you have cleaned up your Applications and Accounts. You cannot delete your service provider if you only have one.

PR #24, opened 2021-08-30
Add account sid selector to speech form

kitajchuk:
This PR:

- Adds account selector to speech form
- Sends `account_sid` with payload if selected in speech form
- Otherwise sends `service_provider_sid` with payload if no account is selected in speech form
- Adds account selector to speech credentials list view
- Adds account selector to carrier list view
- Adds file input pseudo element styling for Mozilla so file input button UI is visible in Firefox

### DIFF
--- a/src/components/elements/FileUpload.js
+++ b/src/components/elements/FileUpload.js
@@ -46,6 +46,38 @@ const StyledInput = styled.input`
   &:active:not([disabled]):after {
     background: #A40D40;
   }
+
+  &::file-selector-button {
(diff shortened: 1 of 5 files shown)


davehorton:
looks good.  Can you make one final change on this branch, and then lets merge and work on the bug fixes on a different branch.

Could you add same "Used by" dropdown selector on the Carrier list view that you now have on the Speech Services list view?

image

kitajchuk:
@davehorton I've pushed a commit that adds the account selector to the Carrier list view, same as Speech service list view. This PR is open for review/merge.

PR #28, opened 2021-10-14
SMS for Carrier Form

kitajchuk:


davehorton:
It's correct as you have it, that the user needs to fill in Inbound SMPP System in the open source version.

Beyond that, a couple of changes needed I think.

1) It looks like smm_inbound_system_id is actually being properly saved in the db, and returned in the GET, but I think the problem is that the webapp is missing a call to setSmppInboundSystemId [at or about here](https://github.com/jambonz/jambonz-webapp/blob/2962855322996baa383a5750ef6648080cf4f2d0/src/components/forms/CarrierForm.js#L424)

2) Since we now have to tell the user about the public IPs for both SIP traffic and SMPP traffic, and they could be different, could you move that information under the Voice and SMS tabs, respectively as done in the hosted version.  See below.

image

image

kitajchuk:
@davehorton Good spot on the setter—thanks. I pushed some more changes. I think my final question, maybe, is what to do about the first line of the help text for the Voice form: "Have your carrier send calls to {some sip realm}"? There seems to be no default SIP realm in the open source version since you can have many accounts. Also you can choose to associate the carrier with a single account (which may or may not have an SIP realm) or with all accounts. Then there is also an option to enter a custom SIP realm when you check the box for "Carrier requires SIP Register before sending outbound calls". At the moment I have it using that last one if present but I really don't know what it should be.

davehorton:
hmm, wait let me think about this...

kitajchuk:
@davehorton That change is in. PR is open now 👍

PR #29, opened 2021-10-18
Update IP whitelist tooltips for carrier form

kitajchuk:
@davehorton I've made the change you requested. The only difference is I'm using the IPs from the `/Sbcs` endpoint for the Voice tab which seem to differ from what you had in the email. The SMPP IPs are the same, however. If we need to use the ones from the email for SIP IPs we can just hard code them instead. Let me know, otherwise this should be good to go 👍


davehorton:
A couple of issues.  First, the Tooltip for Voice references SMPP (voice=SIP) and the ToolTip for SMS references SIP (sms=smpp) so those need to be swapped.

image

image

The other thing is that the tooltip for SMPP needs an additional bit of information, in the case where we are expecting TLS connections (noted by the use_tls property returned in the /Smpps api call).  In that case, the tooltip should read
```
Have your carriers send SIP calls to our servers at:
3.209.58.102:3550 (TLS)
34.197.99.29:2775
```

PR #32, opened 2021-11-15
Add Microsoft vendor to Jambonz Webapp

kitajchuk:


kitajchuk:
@davehorton Second commit adds the MS vendor to the Application form. I believe I've got the data correct but please verify that everything looks correct for these changes please 👍

davehorton:
I think the functionality looks good, but here are a few notes on the UI:

2) The prompt about signaling IPs now show up on the Speech credentials page:

image

2) When adding a speech credential, if you select Microsoft you no longer get the checkboxes "use for tts", "use for stt".  This is probably why the credentials never fail, because we never test on the server side since the creds are not enabled for either service.

image

3) On the Add/Edit Application page, if you select Microsoft as Speech Synthesis vendor, it seems to punch the voice dropdown out through the right edge of the page

image

kitajchuk:
@davehorton I fixed those 3 things. I squashed this branch so as long as things look good I think we're good 👍

davehorton:
Looks good.  Only one issue left -- the speech form is still showing the text that should only appear on the Carrier panel

image

kitajchuk:
@davehorton Okay—pushed a commit to fix that. I guess I had removed the `Sbcs` copy from the Speech Services list page but I missed it on the actual form—doh!

PR #35, opened 2021-12-06
Subspace

kitajchuk:


kitajchuk:
@davehorton I've got this sending the `destination` payload as per the document specification. As you mentioned I do get a 500 server error when posting so I figure you need to pull this and wrap up the server side of things. Let me know if this client implementation needs any tweaks 👍

davehorton:
First of all, I've pushed changes to the server so now you should be able to fully test.

The 'Enable' function works almost perfectly now.  The only thing is that once we enable Subspace, we would like to show some information to the user on the UI about the enabled "pipe".  Specifically, we want to show them the other end of the pipe, since they will want to have their endpoints send traffic there.

To help that, I am sending back a 200 to the POST with this payload
```
{
	"subspace_sip_teleport_id": "8eecddcc-3027-4d0a-8267-afa625a5452e",
	"subspace_sip_teleport_destinations": [{
		"address": "129.203.31.5:5092",
		"transport_type": "UDP_TCP"
	}, {
		"address": "129.203.31.5:10092",
		"transport_type": "TLS"
	}]
}
```
By the way, these values are also stored in the back-end db in accounts.subspace_sip_teleport_id, and accounts.subspace_sip_teleport_destinations (where we store a JSON-ifyed string of the property of same name above).

I would like to have this displayed on the UI, probably in 3 lines, like this:
```
Subspace is now enabled.  To send your traffic through Subspace:
- send UDP and TCP traffic to 129.203.31.5:5092
- send TLS traffic to 129.203.31.5:10092
```

The 'Disable' function has an issue in that it also pops up the same popup dialog that you implemented for Enable.  For Disable, I think it would be nice to have a confirmation popup ("Are you sure you want to delete your Subspace SIP Teleport?")

delete should be simply DELETE Accounts/:accountSid/SubspaceTeleport

davehorton:
Oh I forgot to mention, I have checked in a change to the jambonz-api-server on the branch you are testing so please get the latest and test with that

kitajchuk:
@davehorton I pushed another update to address these issues. It seems I still cannot fully test all functions because when I try to `DELETE` I get an "Unauthorized" response. When I try to add the Subspace Teleport to another account I create from scratch I get a response of "Too Many Requests". Perhaps this is because I'm using the same Subspace client ID and Secret for each account?

davehorton:
OK, I pushed some changes to the backend.  Delete now works, mostly.  Could you grab the latest and make these fixes?

1. Delete works, but the UI continues to show the new text below about the (now-deleted) subspace artifact.  Upon a successful delete (204 response) can you update that UI to remove that 
2. If you try to add a second subspace "pipe" you will hit a limit on subspace due to the fact that my account is only allowed a max of one.  The response to the webapp is a 500 status with a JSON payload "{"msg":"Too Many Requests"}".  In this case, could you handle this and show an error message "You have already created the maximum number of SIP Teleports allowed for your Subspace account".
3. UI thing -- the new text looks good but could you reduce the line spacing between line 1-2 and line 2-3

image

kitajchuk:
@davehorton Those changes are pushed—go ahead and test them out 👍

davehorton:
Awesome!  great work

PR #37, opened 2022-01-13
update dependencies and fix security vulnerabilities

akirilyuk:
This PR fixes almost all current security vulnerabilities and also fixes some linting issues because of which react could not  compile.

I have verified the functionality by trying out every view in the UI and doing some CRUD operations with a remote running Jambonz API server.


akirilyuk:
Before:
Screenshot 2022-01-13 at 17 26 33
After:
Screenshot 2022-01-13 at 17 27 13

PR #48, opened 2022-04-16
webpack issue fix

cluzier:
Cleaned up terminal on web-app start:
- When running, it no longer logs `Failed to parse source map: ... URL is not supported`


davehorton:
I actually still see webpack errors (actually, more of them) when I test these changes.  See [this console log](https://gist.github.com/davehorton/a126a80a82cffa9b50a0eaa243fc16ae)

PR #50, opened 2022-04-18
regions field added to aws, default to us-east-1

### LINKED ISSUE #49: Add region dropdown when creating an AWS speech credential
Currently, when adding a speech credential for AWS the user is prompted only for access key and secret access key (see below).

image

An additional dropdown for should appear below the "secret access key" field with a Label "Region:".  The dropdown values can be found [in this file](https://github.com/jambonz/jambonz-webapp/blob/main/src/data/AwsRegions.js).

When adding a new speech credential, the region dropdown should initially default to us-east-1, but the user can then change it to any of the available regions before saving.

cluzier:
fixes #49 

- region field added to aws Speech Service
- defaults to us-east-1


davehorton:
after collecting the region, the POST request must include a new property named 'aws_region' with the value of the selected region.

davehorton:
Better, but next problem: when retrieving the credential the wrong region is displayed -- even though 'us-west-1' is retrieved the gui shows 'us-east-1'

image

davehorton:
the change to require a region when creating an aws credential worked.  However, when retrieving the credential I just created in us-west-2, the region is not shown.  Instead, I see "select a region".  The aws region associated with the credential should be selected in the dropdown

image

sammachin:
Would this not be better as a text box than a dropdown? With a dropdown it will need to be updated everytime AWS adds a new region.

cluzier:
Still trying to figure out why the dropdown isn't showing the last region the user picked. No longer showing the default "Select a region" on the edit Speech Service page

davehorton:
I found and fixed an error on the server side related to editing an existing speech credential, seems to work properly now.

PR #53, opened 2022-04-28
removed radio buttons, replaced with dropdown

### LINKED ISSUE #52: Change vendor list in Speech Service to dropdown
The list of vendors in Add/Edit Speech Service is currently represented by radio buttons:
image

As we add more vendors this will become unwieldly.  The vendor selection should be changed to a dropdown.

cluzier:
Removed vendor radio buttons and replaced with a dropdown menu

fixes #52


davehorton:
could you fix a couple of things:

1.  If I click Add Speech Service without selecting a vendor, I should get a more meaningful error message - or perhaps the Add button should be disabled if a vendor is not selected

image

2. I would like to move the 2 checkboxes up so they vertically come after the "Used by" dropdown.

image

Issue #57, opened 2022-05-23
Issues adding trunk and phone number to service provider with no accounts


davehorton:
Steps to recreate:

1. create a new service provider (SP) in the portal
2. create a new carrier for that service provider

**Problem**: after saving the carrier, the carrier list is empty.  It should show the carrier you just created

Now, if you did the steps this way, the problem does not occur:

1. create a new service provider in the portal
2. create a new account within that service provider
3. create a new carrier for that service provider

Then the problem does not happen.  In other words, if you immediately create an account after creating a service provider, you do not have the issue.  But it should not be required to do things in this order: you should be able to create an SP and then a Carrier for that SP without first creating an account

khanghugo:
This issue comes from this one line.
https://github.com/jambonz/jambonz-webapp/blob/bb335d08387d20706ec0308f3cfd4cf098563ea1/src/components/pages/internal/CarriersList.js#L107

Simply removing it will fix this problem.

However, I am not sure why it is there at the first place. There should be a reason why it is there at the first place. So there is something about it that we can work toward it.

khanghugo:
From this, I also discovered that this line is called twice when there is no accounts but once when there is an account.
https://github.com/jambonz/jambonz-webapp/blob/bb335d08387d20706ec0308f3cfd4cf098563ea1/src/components/pages/internal/CarriersList.js#L95

It is also safe to assume that some other methods on higher levels are also repeating. That seems like somewhere to fix also, maybe a bit minor.

PR #58, opened 2022-05-23
Fix issue #57

khanghugo:


davehorton:
OK that looks good.

However, there is another issue I would like you to fix.

Now, I can add a new SP and then I can add a Carrier and see it under that SP.  That's great.

However, if right after that, I add a new phone number under that SP, associate it with the new SP, and leave the account blank (because I have not added an account under that SP), the phone number gets added under the 'default' SP.

Could you look into fixing that as well, and push the changes to this PR

khanghugo:
Hmmm. I see. So what I think might be the fix is to create a new account for that phone number when there is no account. If I do that then should there also be an option to create a new account associated with the phone number in the phone number tab. 

On the other hand, we can create a default account for a new SP and that phone number will be under that default account.

This is an edge case that I think create a default account is the most foolproof.

davehorton:
No, please do not create a new account.

If you look at the POST request with the new phone number, you will the problem is that for some reason it is including an account sid, even though you did not specify one

You should be able to add a phone number as long as you have a carrier.  Associating a phone number to an account is purely optional

khanghugo:
I see where it is now.
https://github.com/jambonz/jambonz-webapp/blob/bb335d08387d20706ec0308f3cfd4cf098563ea1/src/components/forms/PhoneNumberForm.js#L51

From here, the API request get all of the available information, including ones from other SP. What I can do here is to filter out the SPs that don's match with the current SP we are looking at.

davehorton (approved):

Issue #59, opened 2022-05-25
Microsoft Azure speech region list needs to be updated


davehorton:
The latest list is [here](https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/regions#speech-to-text-pronunciation-assessment-text-to-speech-and-translation)

We should update [this code](https://github.com/jambonz/jambonz-webapp/blob/main/src/data/AwsRegions.js) accordingly.

PR #60, opened 2022-05-25
Update Azure speech list

khanghugo:
For #59

### DIFF
--- a/src/data/MicrosoftAzureRegions.js
+++ b/src/data/MicrosoftAzureRegions.js
@@ -1,96 +1,94 @@
 const regions = [
-  {
-    name: 'Asia (East)',
-    value: 'eastasia'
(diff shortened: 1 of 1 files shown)


davehorton (changes requested):
changes requested:
1. Change US regions from for instance "East US" to "US - East".  
2. Make the dropdown list appear in alphabetical order
2. Change each menu item from, for instance, "US - East" to "US - East (eastus)", ie put the internal azure identifier for each region in parentheses after the display name

image

khanghugo:
Okay I see adding the hyphen and sorting makes it look prettier. But should there be the identifier? Would the drop down menu look too wide? Let me do that and see.

davehorton (approved):

PR #63, opened 2022-07-09
Add z-index property to Add button

juansalvatella:
The add button on the internal template can't be clicked because it's under the list component. I've added a z-index CSS property to send it to the top. Doing so the add button can be clicked anywhere down the list and not just at the beggining.

### DIFF
--- a/src/components/elements/AddButton.js
+++ b/src/components/elements/AddButton.js
@@ -21,6 +21,7 @@ const StyledLink = styled(FilteredLink)`
   border-radius: 50%;
   text-decoration: none;
   color: #565656;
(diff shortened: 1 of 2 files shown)


khanghugo:
Hey, thanks for the contribution. 

I am personally unable to replicate a problem involving the z-index to test this change. If possible, can you include a screenshot where it becomes an issue? I can merge it right away.

juansalvatella:
Hi there, you should have a list long enough so that the "Add" button is on top of the icon to act on each item. As you can see on this screenshot:
image

davehorton (approved):

PR #64, opened 2022-07-12
Webapp refresh!

kitajchuk:
This is a WIP so opening a Draft for now so we can track...


khanghugo:
Hey, I have been looking around the thing and I really like it. Lots of things got chopped down to smaller pieces so it is easier to look around. Lots of (self) documentation too so that will help me a lot with on boarding this.

kitajchuk:
> Hey, I have been looking around the thing and I really like it. Lots of things got chopped down to smaller pieces so it is easier to look around. Lots of (self) documentation too so that will help me a lot with on boarding this.

Awesome! Glad to hear that. I definitely have put a lot into making sure this is the case! I considered jsdocs as a requirement via linting but decided against it since it’s a burden on developers and ideally we write the code well enough it can be understood easily. I think typescript is also helping in this sense to articulate code clarity. I think longer term having test coverage would be good but also 🤔. For next steps I’m going to create issues on this repo for the remaining work with acceptance criteria based on parity of function. From there we can work out where it makes the most sense to start for you but also feel free to just dive in where you want once we have issues to tackle 🤘

kitajchuk:
@khanghugo I was just playing around and found a runtime error on the applications view when you don't yet have any accounts for the current service provider. I tested on the old webapp and they just redirect you to accounts with a toast when you try to create an application but there are no accounts. I thought we could handle that redirect scenario as well as a last ditch fallback but also provide more up front UI to help the user understand they need to first create an account. This way you are not surprised by a redirect because the applications view provides messaging directing you to go create an account. Commit here: https://github.com/jambonz/jambonz-webapp/pull/64/commits/2201644df11e20ba34abc175196c27d480fa62bc

khanghugo:
Woo-hoo

Issue #66, opened 2022-08-03
Add support for setting a SIPREC application at the account level


davehorton:
We want to make change to the Add/Edit account panel to allow setting a new variable.  Currently, the panel looks like this:

![image](https://user-images.githubusercontent.com/1884973/182626897-158554af-8d1d-4816-b673-edd0aa01f15b.png)

What we want to do is add a new field right below "Application for SIP Device Calls".

In fact, we want it to be very similar.  It will be a field just like that one, which will show the same dropdown list of applications and let the user choose one.  The label should read "Application for SIPREC Calls"

In the POST or PUT to the API, include the application sid identifier in a property named "siprec_hook_sid"

davehorton:
Also can yo make this UI change?  Right align the dropdown for "Application for SIP Device Calls" and then put the label on one line instead of two lines.  And please do the same with the new label and dropdown

khanghugo:
Got it.

PR #67, opened 2022-08-03
Implementing SIPREC application selection (#66)

khanghugo:
I want to make sure the variables look right first before making any stylistic changes.

As for stylistic changes, it is embarrasing to admit that I was spending the last hours to align text. There is more into it. The form is a grid because it has the attribute `large`. So, in order for the text to stay in one line, we either tread the on the grid form or have more column space for text to render (this does not change the content box side, yet). This will affect other labels in this page. But it only works for accounted render resolution. So on some devices, it will become two lines. So we can do something like render the form at device's resolution, something like that. But that would change a bit more things. 

Anyway, here is what it looks like.

image

If the variable looks okay then I can try to make it look a bit fancier as requested.

### DIFF
--- a/src/components/forms/AccountForm.js
+++ b/src/components/forms/AccountForm.js
@@ -73,6 +73,7 @@ const AccountForm = props => {
   const [ name,          setName       ] = useState('');
   const [ sipRealm,      setSipRealm   ] = useState('');
   const [ deviceCallingApplication, setDeviceCallingApplication ] = useState('');
+  const [ siprecCallingApplication, setSiprecCallingApplication ] = useState('');
   const [ regWebhook,    setRegWebhook ] = useState('');
   const [ regMethod,     setRegMethod     ] = useState('POST');
   const [ regUser,       setRegUser       ] = useState('');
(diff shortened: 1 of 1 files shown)


davehorton:
This looks good.  Let's not worry about the UI changes I requested for right now.  Let me review this and if it looks good I will merge.  Thanks

davehorton:
Looks great!

davehorton:
oops, meant to merge but I closed it..

src/components/forms/AccountForm.js
```diff
@@ -695,33 +698,64 @@ const AccountForm = props => {
...
+                ))}
+              </Select>
+            </React.Fragment>
+            
+            <React.Fragment>
+              <Label tooltip htmlFor="siprecCallingApplication">
+                <span style={{ position: 'relative' }}>
+                  Application for SIPREC Calls
+                  <Tooltip large>
+                    This is a really long sample text to let it show up something
```
kitajchuk, line 735: Hey looks like this dummy text got merged to `main` -- do we have real text for this or should we just omit the tooltip here?

khanghugo: Oh yea, I wasn't sure what to write there... I think omission should be fine. It would be similar to the other one.

kitajchuk: You wanna put up a PR to do that then?

khanghugo: Okay, let me make one.

kitajchuk: Cool. Also in the meantime take a look at the `refresh` branch. I’ll be onboarding you to that here soon to work on the rest of the views. It’s a complete rewrite of this application and I’m pretty stoked on how it’s shaped up 🙂

khanghugo: Nice job. I've been waiting for this.

kitajchuk: Awesome! Might also be good to check out the core `jambonz-ui` package 📦. This is where I’ve extracted the reusable design elements for building jambonz branded frontends. I think it’s possible some of the reusable components from the web app refresh are going to be moved into the ui library 🤔.

PR #69, opened 2022-08-05
remove dummy text from siprec tooltip

khanghugo:
Not sure why the other one just automatically close, probably I pushed before making a fix commit


kitajchuk (approved):

kitajchuk:
> Not sure why the other one just automatically close, probably I pushed before making a fix commit

Looks good 👍 go ahead and merge it

PR #77, opened 2022-08-12
Application page for refresh branch

khanghugo:
#70 

For now, it is barely working. I know why but I can't really bring up the solution for that. My hypothesis is I have no idea how to do async so the pulled array is empty, aka not waiting for the promise. 

I am not sure what is the more correct way to do those things so I do need some guidance on this.

Besides that, I have tested with some hacks and it works with the backend nicely, as far as checking whether the entries get in, get out, and get deleted.

Issue #70, opened 2022-08-10
Refresh: Applications view


kitajchuk:
Implement functional parity for the `Applications` view form with the new UI style.

Create a branch from `refresh` called `refresh-applications` and submit a PR to the `refresh` branch. Please include this issue number (`#70`) in the commit/PR message.

You can find the container stub for this view [here](https://github.com/jambonz/jambonz-webapp/tree/refresh/src/containers/internal/views/applications). It already has the `vendor` select menu logic implemented which can be used for the form. Review the `Accounts` view implementation [here](https://github.com/jambonz/jambonz-webapp/tree/refresh/src/containers/internal/views/accounts) to get an idea of the new approach to implementing views a la the `list/add/edit` model.

- [ ] The add/edit flow needs to have functional parity with current web app
- [ ] The list view also needs to include the accounts filter functionality
- [ ] When deleting an application you must first verify that the app is not being used by an Account or MS Teams Tenant

New routes to be added:

- `/internal/applications/add`
- `/internal/applications/application_sid/edit`

For any questions about design/UI/coding ping @kitajchuk.

Ultimately, when merging PRs, please always use "Squash and Merge" for the strategy.

kitajchuk:
@khanghugo I've updated the issue information based on our discussions on the current PR review.

PR #79, opened 2022-08-12
Application page for refresh 

khanghugo:
#70

Some more original writing from the previous post #77 .

It is barely working because I am not sure how I should use, or maybe should do, the async stuff because data returned from the server is empty until that happens. And right now, it is still empty. But I have made sure that everything has to work beside that.

One small problem is I was really trying to fix the compiler error so maybe there is some weird thing there but let's continue after this.

**Checklist for completion:**

- [x] The add/edit flow needs to have full functional parity with current web app
- [x] The list view also needs to include the accounts filter functionality using new `AccountFilter` component
- [x] When deleting an application you must first verify that the app is not being used by an Account or MS Teams Tenant

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -36,7 +36,6 @@ import type {
   Alert,
   PagedResponse,
   RecentCall,
-  Application,
 } from "./types";
 
 /** Wrap all requests to normalize response handling */
@@ -281,10 +280,6 @@ export const getAccountWebhook = (sid: string) => {
   return getFetch<SecretResponse>(
     `${API_ACCOUNTS}/${sid}/WebhookSecret?regenerate=true`
   );
-};
-
-export const getApplications = (sid: string) => {
-  return getFetch<Application[]>(`${API_ACCOUNTS}/${sid}/Applications`);
 };
 
 /** Wrappers for APIs that can have a mock dev server response */
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -10,6 +10,7 @@ import {
   API_SERVICE_PROVIDERS,
   API_API_KEYS,
   API_ACCOUNTS,
+  API_APPLICATIONS,
 } from "./constants";
 import { ROUTE_LOGIN } from "src/router/routes";
 import {
@@ -35,6 +36,7 @@ import type {
   Alert,
   PagedResponse,
   RecentCall,
+  Application,
 } from "./types";
 
 /** Wrap all requests to normalize response handling */
@@ -200,60 +202,112 @@ export const postLogin = (payload: Payload) => {
   });
 };
 
-/** Named wrappers for `postFetch` */
-
 export const postServiceProviders = (payload: Payload) => {
-  return postFetch<SidResponse>(API_SERVICE_PROVIDERS, payload);
+  return fetchTransport<SidResponse>(API_SERVICE_PROVIDERS, {
+    method: "POST",
+    body: JSON.stringify(payload),
+    headers: getAuthHeaders(),
+  });
 };
 
 export const postApiKey = (payload: Payload) => {
-  return postFetch<TokenResponse>(API_API_KEYS, payload);
+  return fetchTransport<TokenResponse>(API_API_KEYS, {
+    method: "POST",
+    body: JSON.stringify(payload),
+    headers: getAuthHeaders(),
+  });
 };
 
 export const postAccount = (payload: Payload) => {
-  return postFetch<SidResponse>(API_ACCOUNTS, payload);
+  return fetchTransport<SidResponse>(API_ACCOUNTS, {
+    method: "POST",
+    body: JSON.stringify(payload),
+    headers: getAuthHeaders(),
+  });
 };
 
 export const postSubspace = (sid: string, payload: Payload) => {
-  return postFetch<SidResponse>(
+  return fetchTransport<SidResponse>(
     `${API_ACCOUNTS}/${sid}/SubspaceTeleport`,
-    payload
+    {
+      method: "POST",
+      body: JSON.stringify(payload),
+      headers: getAuthHeaders(),
+    }
   );
 };
 
-/** Named wrappers for `putFetch` */
+export const postApplication = (payload: Payload) => {
+  return fetchTransport<SidResponse>(API_APPLICATIONS, {
+    method: "POST",
+    body: JSON.stringify(payload),
+    headers: getAuthHeaders(),
+  });
+};
 
 export const putUser = (sid: string, payload: Payload) => {
-  return putFetch<EmptyResponse>(`${API_USERS}/${sid}`, payload);
+  return fetchTransport<EmptyResponse>(`${API_USERS}/${sid}`, {
+    method: "PUT",
+    body: JSON.stringify(payload),
+    headers: getAuthHeaders(),
+  });
 };
 
 export const putServiceProvider = (sid: string, payload: Payload) => {
-  return putFetch<EmptyResponse>(`${API_SERVICE_PROVIDERS}/${sid}`, payload);
+  return fetchTransport<EmptyResponse>(`${API_SERVICE_PROVIDERS}/${sid}`, {
+    method: "PUT",
+    body: JSON.stringify(payload),
+    headers: getAuthHeaders(),
+  });
 };
 
 export const putAccount = (sid: string, payload: Payload) => {
-  return putFetch<EmptyResponse>(`${API_ACCOUNTS}/${sid}`, payload);
+  return fetchTransport<EmptyResponse>(`${API_ACCOUNTS}/${sid}`, {
+    method: "PUT",
+    body: JSON.stringify(payload),
+    headers: getAuthHeaders(),
+  });
 };
 
-/** Named wrappers for `deleteFetch` */
-
 export const deleteServiceProvider = (sid: string) => {
-  return deleteFetch<EmptyResponse>(`${API_SERVICE_PROVIDERS}/${sid}`);
+  return fetchTransport<EmptyResponse>(`${API_SERVICE_PROVIDERS}/${sid}`, {
+    method: "DELETE",
+    headers: getAuthHeaders(),
+  });
 };
 
 export const deleteApiKey = (sid: string) => {
-  return deleteFetch<EmptyResponse>(`${API_API_KEYS}/${sid}`);
+  return fetchTransport<EmptyResponse>(`${API_API_KEYS}/${sid}`, {
+    method: "DELETE",
+    headers: getAuthHeaders(),
+  });
 };
 
 export const deleteAccount = (sid: string) => {
-  return deleteFetch<EmptyResponse>(`${API_ACCOUNTS}/${sid}`);
+  return fetchTransport<EmptyResponse>(`${API_ACCOUNTS}/${sid}`, {
+    method: "DELETE",
+    headers: getAuthHeaders(),
+  });
 };
 
 export const deleteSubspace = (sid: string) => {
-  return deleteFetch<EmptyResponse>(`${API_ACCOUNTS}/${sid}/SubspaceTeleport`);
+  return fetchTransport<EmptyResponse>(
+    `${API_ACCOUNTS}/${sid}/SubspaceTeleport`,
+    {
+      method: "DELETE",
+      headers: getAuthHeaders(),
+    }
+  );
 };
 
-/** Named wrappers for `getFetch` */
+export const deleteApplication = (sid: string) => {
+  return fetchTransport<EmptyResponse>(`${API_APPLICATIONS}/${sid}`, {
+    method: "DELETE",
+    headers: getAuthHeaders(),
+  });
+};
+
+/** Specific use cases for wrapping `getFetch` */
 
 export const getUser = (sid: string) => {
   return getFetch<User>(`${API_USERS}/${sid}`);
@@ -267,6 +321,10 @@ export const getAccountWebhook = (sid: string) => {
   return getFetch<SecretResponse>(
     `${API_ACCOUNTS}/${sid}/WebhookSecret?regenerate=true`
   );
+};
+
+export const getApplications = () => {
+  return getFetch<Application[]>(API_APPLICATIONS);
 };
 
 /** Wrappers for APIs that can have a mock dev server response */
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -241,6 +241,10 @@ export const putAccount = (sid: string, payload: Payload) => {
   return putFetch<EmptyResponse>(`${API_ACCOUNTS}/${sid}`, payload);
 };
 
+export const putApplication = (sid: string, payload: Payload) => {
+  return putFetch<EmptyResponse>(`${API_APPLICATIONS}/${sid}`, payload);
+};
+
 /** Named wrappers for `deleteFetch` */
 
 export const deleteServiceProvider = (sid: string) => {
@@ -277,10 +281,6 @@ export const getAccountWebhook = (sid: string) => {
   return getFetch<SecretResponse>(
     `${API_ACCOUNTS}/${sid}/WebhookSecret?regenerate=true`
   );
-};
-
-export const getApplications = () => {
-  return getFetch<Application[]>(API_APPLICATIONS);
 };
 
 /** Wrappers for APIs that can have a mock dev server response */
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -283,6 +283,10 @@ export const getAccountWebhook = (sid: string) => {
   );
 };
 
+export const getApplications = (sid: string) => {
+  return getFetch<Application[]>(`${API_ACCOUNTS}/${sid}/Applications`);
+};
+
 /** Wrappers for APIs that can have a mock dev server response */
 
 export const getRecentCalls = (sid: string) => {
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -202,112 +202,68 @@ export const postLogin = (payload: Payload) => {
   });
 };
 
+/** Named wrappers for `postFetch` */
+
 export const postServiceProviders = (payload: Payload) => {
-  return fetchTransport<SidResponse>(API_SERVICE_PROVIDERS, {
-    method: "POST",
-    body: JSON.stringify(payload),
-    headers: getAuthHeaders(),
-  });
+  return postFetch<SidResponse>(API_SERVICE_PROVIDERS, payload);
 };
 
 export const postApiKey = (payload: Payload) => {
-  return fetchTransport<TokenResponse>(API_API_KEYS, {
-    method: "POST",
-    body: JSON.stringify(payload),
-    headers: getAuthHeaders(),
-  });
+  return postFetch<TokenResponse>(API_API_KEYS, payload);
 };
 
 export const postAccount = (payload: Payload) => {
-  return fetchTransport<SidResponse>(API_ACCOUNTS, {
-    method: "POST",
-    body: JSON.stringify(payload),
-    headers: getAuthHeaders(),
-  });
+  return postFetch<SidResponse>(API_ACCOUNTS, payload);
 };
 
 export const postSubspace = (sid: string, payload: Payload) => {
-  return fetchTransport<SidResponse>(
+  return postFetch<SidResponse>(
     `${API_ACCOUNTS}/${sid}/SubspaceTeleport`,
-    {
-      method: "POST",
-      body: JSON.stringify(payload),
-      headers: getAuthHeaders(),
-    }
+    payload
   );
 };
 
 export const postApplication = (payload: Payload) => {
-  return fetchTransport<SidResponse>(API_APPLICATIONS, {
-    method: "POST",
-    body: JSON.stringify(payload),
-    headers: getAuthHeaders(),
-  });
+  return postFetch<SidResponse>(API_APPLICATIONS, payload);
 };
 
+/** Named wrappers for `putFetch` */
+
 export const putUser = (sid: string, payload: Payload) => {
-  return fetchTransport<EmptyResponse>(`${API_USERS}/${sid}`, {
-    method: "PUT",
-    body: JSON.stringify(payload),
-    headers: getAuthHeaders(),
-  });
+  return putFetch<EmptyResponse>(`${API_USERS}/${sid}`, payload);
 };
 
 export const putServiceProvider = (sid: string, payload: Payload) => {
-  return fetchTransport<EmptyResponse>(`${API_SERVICE_PROVIDERS}/${sid}`, {
-    method: "PUT",
-    body: JSON.stringify(payload),
-    headers: getAuthHeaders(),
-  });
+  return putFetch<EmptyResponse>(`${API_SERVICE_PROVIDERS}/${sid}`, payload);
 };
 
 export const putAccount = (sid: string, payload: Payload) => {
-  return fetchTransport<EmptyResponse>(`${API_ACCOUNTS}/${sid}`, {
-    method: "PUT",
-    body: JSON.stringify(payload),
-    headers: getAuthHeaders(),
-  });
+  return putFetch<EmptyResponse>(`${API_ACCOUNTS}/${sid}`, payload);
 };
 
+/** Named wrappers for `deleteFetch` */
+
 export const deleteServiceProvider = (sid: string) => {
-  return fetchTransport<EmptyResponse>(`${API_SERVICE_PROVIDERS}/${sid}`, {
-    method: "DELETE",
-    headers: getAuthHeaders(),
-  });
+  return deleteFetch<EmptyResponse>(`${API_SERVICE_PROVIDERS}/${sid}`);
 };
 
 export const deleteApiKey = (sid: string) => {
-  return fetchTransport<EmptyResponse>(`${API_API_KEYS}/${sid}`, {
-    method: "DELETE",
-    headers: getAuthHeaders(),
-  });
+  return deleteFetch<EmptyResponse>(`${API_API_KEYS}/${sid}`);
 };
 
 export const deleteAccount = (sid: string) => {
-  return fetchTransport<EmptyResponse>(`${API_ACCOUNTS}/${sid}`, {
-    method: "DELETE",
-    headers: getAuthHeaders(),
-  });
+  return deleteFetch<EmptyResponse>(`${API_ACCOUNTS}/${sid}`);
 };
 
 export const deleteSubspace = (sid: string) => {
-  return fetchTransport<EmptyResponse>(
-    `${API_ACCOUNTS}/${sid}/SubspaceTeleport`,
-    {
-      method: "DELETE",
-      headers: getAuthHeaders(),
-    }
-  );
+  return deleteFetch<EmptyResponse>(`${API_ACCOUNTS}/${sid}/SubspaceTeleport`);
 };
 
 export const deleteApplication = (sid: string) => {
-  return fetchTransport<EmptyResponse>(`${API_APPLICATIONS}/${sid}`, {
-    method: "DELETE",
-    headers: getAuthHeaders(),
-  });
+  return deleteFetch<EmptyResponse>(`${API_APPLICATIONS}/${sid}`);
 };
 
-/** Specific use cases for wrapping `getFetch` */
+/** Named wrappers for `getFetch` */
 
 export const getUser = (sid: string) => {
   return getFetch<User>(`${API_USERS}/${sid}`);
--- a/src/containers/internal/views/applications/delete.tsx
+++ b/src/containers/internal/views/applications/delete.tsx
@@ -88,7 +88,7 @@ export const DeleteApplication = ({
     return function cleanup() {
       ignore = true;
     };
-  });
+  }, []);
 
   return (
     <>
--- a/src/containers/internal/views/applications/delete.tsx
+++ b/src/containers/internal/views/applications/delete.tsx
@@ -1,0 +1,41 @@
+import React from "react";
+import { P } from "jambonz-ui";
+
+import { Modal } from "src/components";
+
+import type { Application } from "src/api/types";
+
+type DeleteProps = {
+  application: Application;
+  handleCancel: () => void;
+  handleSubmit: () => void;
+};
+
+export const DeleteApplication = ({
+  application,
+  handleCancel,
+  handleSubmit,
+}: DeleteProps) => {
+  return (
+    <>
+      <Modal handleCancel={handleCancel} handleSubmit={handleSubmit}>
+        <P>
+          Are you sure you want to delete the application{" "}
+          <strong>{application.name}</strong>?
+        </P>
+        <table>
+          <tr>
+            <th>Account:</th>
+            <th>{application.account_sid}</th>
+          </tr>
+          <tr>
+            <th>Calling webhook:</th>
+            <th>{application.call_hook?.url || "[None]"}</th>
+          </tr>
+        </table>
+      </Modal>
+    </>
+  );
+};
+
+export default DeleteApplication;
--- a/src/containers/internal/views/applications/delete.tsx
+++ b/src/containers/internal/views/applications/delete.tsx
@@ -1,9 +1,11 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { P } from "jambonz-ui";
 
 import { Modal } from "src/components";
+import { useApiData } from "src/api";
+import { toastError } from "src/store";
 
-import type { Application } from "src/api/types";
+import type { Application, Account } from "src/api/types";
 
 type DeleteProps = {
   application: Application;
@@ -11,11 +13,51 @@ type DeleteProps = {
   handleSubmit: () => void;
 };
 
+type DeleteInfoParameter = {
+  item: Application | Account;
+  key1: string;
+  key2?: string;
+  label: string;
+};
+
+const DeleteInfo = ({ item, key1, key2, label }: DeleteInfoParameter) => {
+  return (
+    <ul className="m">
+      <li>
+        <strong>{label}:</strong>
+      </li>
+      <li className="txt--teal">
+        {
+          // this is getting out of hand
+          (item &&
+            ((item[key1 as keyof typeof item] &&
+              key2 &&
+              item[key1 as keyof typeof item]![
+                key2 as keyof typeof item[keyof typeof item]
+              ]) ||
+              item[key1 as keyof typeof item])) ||
+            "[None]"
+        }
+      </li>
+    </ul>
+  );
+};
+
 export const DeleteApplication = ({
   application,
   handleCancel,
   handleSubmit,
 }: DeleteProps) => {
+  const [account, , error] = useApiData<Account>(
+    `Accounts/${application.account_sid}`
+  );
+
+  useEffect(() => {
+    if (error) {
+      toastError(error.msg);
+    }
+  }, [error]);
+
   return (
     <>
       <Modal handleCancel={handleCancel} handleSubmit={handleSubmit}>
@@ -23,16 +65,25 @@ export const DeleteApplication = ({
           Are you sure you want to delete the application{" "}
           <strong>{application.name}</strong>?
         </P>
-        <table>
-          <tr>
-            <th>Account:</th>
-            <th>{application.account_sid}</th>
-          </tr>
-          <tr>
-            <th>Calling webhook:</th>
-            <th>{application.call_hook?.url || "[None]"}</th>
-          </tr>
-        </table>
+        {account && <DeleteInfo item={account} key1="name" label="Account" />}
+        <DeleteInfo
+          item={application}
+          key1="call_hook"
+          key2="webhook_sid"
+          label="Calling Webhook"
+        />
+        <DeleteInfo
+          item={application}
+          key1="call_status_hook"
+          key2="webhook_sid"
+          label="Call Status Webhook"
+        />
+        <DeleteInfo
+          item={application}
+          key1="messaging_hook"
+          key2="webhook_sid"
+          label="Messaging Webhook"
+        />
       </Modal>
     </>
   );
--- a/src/containers/internal/views/applications/delete.tsx
+++ b/src/containers/internal/views/applications/delete.tsx
@@ -1,17 +1,30 @@
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { P } from "jambonz-ui";
 
-import { Modal } from "src/components";
-import { useApiData } from "src/api";
+import { Modal, ModalClose } from "src/components";
+import { useApiData, getFetch } from "src/api";
 import { toastError } from "src/store";
 
-import type { Application, Account } from "src/api/types";
+import type { Application, Account, MSTeamsTenant } from "src/api/types";
+import { API_ACCOUNTS, API_MS_TEAMS_TENANTS } from "src/api/constants";
 
 type DeleteProps = {
   application: Application;
   handleCancel: () => void;
   handleSubmit: () => void;
 };
+
+type InUseProps = {
+  items: Account[] | MSTeamsTenant[];
+  sidKey: string;
+  labelKey: string;
+  itemsLabel: string;
+};
+
+interface InUse {
+  accounts: Account[];
+  teams: MSTeamsTenant[];
+}
 
 const DeleteInfo = ({
   label,
@@ -30,6 +43,23 @@ const DeleteInfo = ({
   );
 };
 
+const InUseItems = ({ items, itemsLabel, sidKey, labelKey }: InUseProps) => {
+  return (
+    <ul className="m">
+      <li>
+        <strong>{itemsLabel}:</strong>
+      </li>
+      {items.map((item) => {
+        return (
+          <li className="txt--teal" key={item[sidKey as keyof typeof item]}>
+            {item[labelKey as keyof typeof item]}
+          </li>
+        );
+      })}
+    </ul>
+  );
+};
+
 export const DeleteApplication = ({
   application,
   handleCancel,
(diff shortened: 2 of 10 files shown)


kitajchuk:
> #70
> 
> Some more original writing from the previous post #77 .
> 
> It is barely working because I am not sure how I should use, or maybe should do, the async stuff because data returned from the server is empty until that happens. And right now, it is still empty. But I have made sure that everything has to work beside that.
> 
> One small problem is I was really trying to fix the compiler error so maybe there is some weird thing there but let's continue after this.

Cool cool. I'll take a look and provide feedback/help :)

khanghugo:
Thanks. Some page will be blank because the return is blank, so no methods like `.map()` for array, thus nothing shows up. But it works if that is.

kitajchuk:
> Thanks. Some page will be blank because the return is blank, so no methods like `.map()` for array, thus nothing shows up. But it works if that is.

Yep—I can point you in the right direction, no worries. Hang tight. I'll submit a review on this PR with notes etc on how to get moving again.

src/api/index.ts
```diff
@@ -245,6 +255,10 @@ export const getAccountWebhook = (sid: string) => {
   );
 };
 
+export const getApplications = () => {
```
kitajchuk, line 258: Let's remove this method. In fact, I intend to remove all `get` methods from the API scope here because we should always just rely on the `use` hooks or calling `getFetch` directly in the few places we may need to. I'll point out in another comment how to use the hooks so you can get rid of this 👍

src/containers/internal/views/applications/delete.tsx
```diff
@@ -0,0 +1,41 @@
...
+  handleSubmit,
+}: DeleteProps) => {
+  return (
+    <>
+      <Modal handleCancel={handleCancel} handleSubmit={handleSubmit}>
+        <P>
+          Are you sure you want to delete the application{" "}
+          <strong>{application.name}</strong>?
+        </P>
+        <table>
```
kitajchuk, line 26: Let's not use this `table` here. If you look at the delete component for the accounts view you can see how I used a UI pattern with a `ul` using the `m` className and a `strong` leading `li` followed by the content. This should be the pattern we use in these modals for consistency and to avoid any tables.

src/containers/internal/views/applications/edit.tsx
```diff
@@ -0,0 +1,39 @@
...
+    if (error) {
+      toastError(error.msg);
+    }
+  }, [error]);
+
+  return (
+    <>
+      <H1>Edit Application</H1>
+      <ApplicationForm application={{ data, refetch, error }} />
+      {/*      <ApiKeys
```
kitajchuk, line 30: You can just remove all this since the applications form doesn't use API keys 👍

src/containers/internal/views/applications/edit.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { useApiData } from "src/api";
+import { toastError } from "src/store";
+import { ApplicationForm } from "./form";
+
+import type { Application } from "src/api/types";
+
+export const EditApplication = () => {
+  // const params = useParams();// TODO get the params correctly and promise? for the data
+  const [data, refetch, error] = useApiData<Application>("Applications"); // return is array but let's just pretend so the compiler stop complaining
```
kitajchuk, line 12: You're really close here. The params are going to be the dynamic values from the route path, in this case it would be `application_sid` which maps to the `:application_sid` from the route path. You can see how this was done in the accounts edit component as well. But what you want looks like this:

```js
const [data, refetch, error] = useApiData<Application>(
  `Applications/${params.application_sid}`
);
```

src/api/index.ts
```diff
@@ -197,6 +199,10 @@ export const postSubspace = (sid: string, payload: Payload) => {
   );
 };
 
+export const postApplication = (payload: Payload) => {
```
kitajchuk, line 202: This is good. You're going to want to also make one for `put` when you get to the edit vs add logic of the form 👍

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,557 @@
...
+};
+
+type ApplicationFormProps = {
+  application?: null | UseApplicationData;
+};
+
+export const ApplicationForm = ({
+  application = null,
+}: ApplicationFormProps) => {
+  // const [application, setApplication] = useState<Application | null>(null);
```
kitajchuk, line 42: You won't need this so just remove.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,557 @@
...
+
+      if (filtered.find((a) => a.name === applicationName)) {
+        setMessage(
+          "The name you have entered is already in use on another one of your applications."
+        );
+        return;
+      }
+    }
+
+    postApplication({
```
kitajchuk, line 176: Here is where you'll conditional out the add vs edit logic for whether you `post` or `put`. Example in the accounts form when you get there 👍

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,557 @@
...
+            />
+          </fieldset>
+          <fieldset>
+            <label htmlFor="account_name">Account</label>
+            <Selector
+              id="account_name"
+              name="account_name"
+              value={accountSid}
+              placeholder="Select an account"
+              // options={[{ // TODO: correctly concating these twos so the select an account option is up there
```
kitajchuk, line 299: Once you have the `application` prop updated (see other comments) you can then use conditional rendering throughout the form here. Reference the accounts form, but the logic is basically this:

```js
{application && application.data && ({
  /*...render things*/
)}
```

You'll use this logic to handle fields render based on whether they are only for add or only for update. Some are for both obviously. And you can use this logic for determining `post` vs `put` as well.

kitajchuk (changes requested):
Great first pass! I added lots of notes and feedback that should get you rolling to bring this one over the finish line 👍

kitajchuk:
@khanghugo if you’re using vscode the extensions config has a list of recommended plugins for linting, prettier and typescript 👍. Definitely make sure you have all these installed for the best dev workflow.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,557 @@
...
+            <>
+              <fieldset>
+                <label htmlFor="recognizer_vendor">
+                  Speech Recognizer Vendor
+                </label>
+                <Selector
+                  id="recognizer_vendor"
+                  name="recognizer_vendor"
+                  value={recogVendor}
+                  options={vendors.slice(0, vendors.length - 1)}
```
kitajchuk, line 496: Hey one thing I was thinking here is that we should use a `filter` rather than this `slice` so that the code would still work even if the vendors were in a different order. Something like:

```js
vendors.filter((vendor) => vendor !== VENDOR_WELLSAID)
```

src/api/index.ts
khanghugo: Funny enough, I have the useApiData to retrieve the data, so this is pretty much unused but the compiler does not complain for some reasons.

kitajchuk: The compiler won’t complain about an unused thing if it’s a module export. If it was not exported and unused within this file it would complain.

src/containers/internal/views/applications/form.tsx
khanghugo: Can I keep this so it looks consistent with the account form? Right now it has the `AccountFormProps` type with just the return of `useApiData()`.

src/containers/internal/views/applications/form.tsx
khanghugo: Yes, this is it. Condtional rendering. This one is new for me because it is way more different from just handling with async stuff (this is nice for me because I won't need to touch more complicated things)

src/containers/internal/views/applications/form.tsx
kitajchuk: Yes keep the props as you have them here. I was referring to the commented out state tuple.

khanghugo: Done deal.

khanghugo:
This looks like it is what I really want now, unless I forgot something or you see something.

I was spending half an hour fixing implicit cast, it is getting tiresome. Compilers are simply my enemies. But at least I have a good laugh out of it.

>if you’re using vscode the extensions config has a list of recommended plugins for linting, prettier and typescript

I was using VSCodium for a while but I cannot get LSP to work on my hobby project (this is entirely my fault as I am using a non standard system). Then I realize that VSCode is unable to find function definitions natively. That is when I just put it off and use something else that just works, which is Sublime Text, and I am comfortable with using ST for simple text editing for a while already. I will definitely give VSCode another chance if those extensions are cool enough.

khanghugo:
https://github.com/jambonz/jambonz-webapp/blob/da0d2fe171606a6ac23283c14bb122d165d77837/src/containers/internal/views/applications/form.tsx#L173-L179

This is not what I want, let me fix that.

kitajchuk:
> This looks like it is what I really want now, unless I forgot something or you see something.
> 
> I was spending half an hour fixing implicit cast, it is getting tiresome. Compilers are simply my enemies. But at least I have a good laugh out of it.
> 
> > if you’re using vscode the extensions config has a list of recommended plugins for linting, prettier and typescript
> 
> I was using VSCodium for a while but I cannot get LSP to work on my hobby project (this is entirely my fault as I am using a non standard system). Then I realize that VSCode is unable to find function definitions natively. That is when I just put it off and use something else that just works, which is Sublime Text, and I am comfortable with using ST for simple text editing for a while already. I will definitely give VSCode another chance if those extensions are cool enough.

Cool cool. I’ve found vscode to be good as long as you install the plugins. Typically it’s good about suggesting them too.

I’ll be able to give you another proper review here maybe tomorrow. If not then Monday for sure. Hang tight. Once this one is in you’ll breeze through the rest 🙂.

How do you like typescript? This is my first time really giving it a proper implementation 🤔.

khanghugo:
> How do you like typescript?

My problem with Typescript right now is I have yet to have a proper IDE for it. It is just a pain for me right now just go back and forth between definition to know what I am dealing with. With that, I am all in for any kind of safety as it enforces me to always check things.



>  This is my first time really giving it a proper implementation thinking.

Your implementations are great. For the most part, I just mirror what you did. There wasn't any problem working with what is given.

kitajchuk:
> > How do you like typescript?
> 
> My problem with Typescript right now is I have yet to have a proper IDE for it. It is just a pain for me right now just go back and forth between definition to know what I am dealing with. With that, I am all in for any kind of safety as it enforces me to always check things.
> 
> > This is my first time really giving it a proper implementation thinking.
> 
> Your implementations are great. For the most part, I just mirror what you did. There wasn't any problem working with what is given.

VS Code is pretty great with typescript 🙂.

src/containers/internal/views/applications/delete.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const DeleteInfo = ({ item, key1, key2, label }: DeleteInfoParameter) => {
+  return (
+    <ul className="m">
+      <li>
+        <strong>{label}:</strong>
+      </li>
+      <li className="txt--teal">
+        {
+          // this is getting out of hand
```
kitajchuk, line 31: I believe in this case you don't want the extra component and abstraction. The reason for that on the account form was due to the nature of having to check 4 different data types. Here you just have 3 fields you want to display. Here you can just code the list and items in the main component and remove this extra complexity.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+  Voice,
+  VoiceLanguage,
+  Language,
+} from "src/vendor/types";
+
+import type { Account, WebHook, Application, FetchError } from "src/api/types";
+
+export type UseApplicationData = {
+  // data fed from ./edit
+  data: Application[] | null; // data returned is an array but there is only 1 (0) that matters
```
kitajchuk, line 37: Don't make this an array then, it should just be `Application | null`.

src/containers/internal/views/applications/edit.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { useApiData } from "src/api";
+import { toastError } from "src/store";
+import { ApplicationForm } from "./form";
+
+import type { Application } from "src/api/types";
+
+export const EditApplication = () => {
+  const params = useParams();
+  const [data, refetch, error] = useApiData<Application[]>(
```
kitajchuk, line 13: This should not be an array type, it should just be `Application`.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+
+type ApplicationFormProps = {
+  application?: null | UseApplicationData;
+};
+
+export const ApplicationForm = ({
+  application = null,
+}: ApplicationFormProps) => {
+  const navigate = useNavigate();
+  const [applicationName, setApplicationName] = useState<string>("");
```
kitajchuk, line 50: You don't need to specify the `string` type here or for any of the other state tuples you've specified `string` for in this file. You can rely on type inference for this so you can remove these 👍

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+          toastError(error.msg);
+        });
+    } else {
+      postApplication({
+        name: applicationName,
+        call_hook: callWebhook || null,
+        account_sid: accountSid || null,
+        messaging_hook: messageWebhook || null,
+        call_status_hook: statusWebhook || null,
+        application_sid: null,
```
kitajchuk, line 216: I think you don't need this `application_sid` null field for the post do you? Then you can declare this payload object ahead of time and use it for both the post and the put to reduce redundancy here 🤔

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+            microsoft: msRecognizer,
+          });
+        }
+      }
+    );
+
+    return function cleanup() {
+      ignore = true;
+    };
+  }, [application]); // makes application loaded
```
kitajchuk, line 309: We can remove unnecessary comments like this since we all know how the `useEffect` deps work

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+              name="application_name"
+              placeholder="Application name"
+              value={applicationName}
+              onChange={(e) => setApplicationName(e.target.value)}
+            />
+          </fieldset>
+          {accounts && (
+            <fieldset>
+              <label htmlFor="account_name">Account</label>
+              <Selector
```
kitajchuk, line 330: Let's get rid of the empty default option for parity with the current form and also add a `required` prop so you can then remove the manual check and message from the validation. We should rely on the browser native `required` functionality wherever possible so we can do less manual validation across the entire app 👍

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+    },
+    {
+      name: "GET",
+      value: "GET",
+    },
+  ];
+
+  const webhooks = [
+    {
+      label: "Call",
```
kitajchuk, line 101: This should be `Calling` for the label.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+      label: "Status",
+      prefix: "status_webhook",
+      stateVal: statusWebhook,
+      stateSet: setStatusWebhook,
+      initialCheck: initialStatusWebhook,
+      refUser: refStatusWebhookUser,
+      refPass: refStatusWebhookPass,
+    },
+    {
+      label: "Message",
```
kitajchuk, line 119: This should be `Messaging` for the label.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+    e.preventDefault();
+
+    setMessage("");
+
+    if (!applicationName) {
+      setMessage("Application name must not be empty");
+      return;
+    }
+
+    if (!accountSid) {
```
kitajchuk, line 152: You should be able to remove this check after implementing my notes on the `Select` component below 👍

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+      });
+    }
+  };
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+
+    setMessage("");
+
+    if (!applicationName) {
```
kitajchuk, line 147: You can remove this check since we are using the browser native `required` attribute. We want to 1) reduce redundancy and 2) reduce manual validation as much as possible for this version of the web app. We can't use this pattern for everything, but we ought to for as much as is possible.

src/containers/internal/views/applications/form.tsx
kitajchuk: This field also needs to be `required` so you probably need to add that to the data here.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+      label: "Call",
+      prefix: "call_webhook",
+      stateVal: callWebhook,
+      stateSet: setCallWebhook,
+      initialCheck: initialCallWebhook,
+      refUser: refCallWebhookUser,
+      refPass: refCallWebhookPass,
+    },
+    {
+      label: "Status",
```
kitajchuk, line 110: This needs to be `Call Status` for the label.

kitajchuk: This field also needs to be `required` so you probably need to add that to the data here.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+                  name={webhook.prefix}
+                  label="Use HTTP Basic Authentication"
+                  initialCheck={webhook.initialCheck}
+                >
+                  <label htmlFor={`${webhook.prefix}_username`}>Username</label>
+                  <input
+                    ref={webhook.refUser}
+                    id={`${webhook.prefix}_username`}
+                    type="text"
+                    name="username"
```
kitajchuk, line 387: Let's use the `webhook.prefix` to name these fields so they are unique. I need to also go back and update the account form to do this as well!

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+                    value={webhook.stateVal?.username || ""}
+                    onChange={(e) =>
+                      handleSetHook(e, webhook.stateVal, webhook.stateSet)
+                    }
+                  />
+                  <label htmlFor={`${webhook.prefix}_password`}>Password</label>
+                  <Passwd
+                    ref={webhook.refPass}
+                    id={`${webhook.prefix}_password`}
+                    name="password"
```
kitajchuk, line 398: Same thing, lets use the `webhook.prefix` for the name here as well.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+  const [applicationName, setApplicationName] = useState<string>("");
+  const [applications] = useApiData<Application[]>("Applications");
+
+  const [accountSid, setAccountSid] = useState<string>("");
+  const [accounts] = useServiceProviderData<Account[]>("Accounts");
+
+  const refCallWebhookUser = useRef<HTMLInputElement>(null);
+  const refCallWebhookPass = useRef<HTMLInputElement>(null);
+  const [callWebhook, setCallWebhook] = useState<WebHook>(DEFAULT_WEBHOOK);
+  const [initialCallWebhook] = useState<boolean>(false);
```
kitajchuk, line 59: Same as with the `string` types you don't need to specify the `boolean` type here. Inference is on our side for this 👍

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+  const refStatusWebhookUser = useRef<HTMLInputElement>(null);
+  const refStatusWebhookPass = useRef<HTMLInputElement>(null);
+  const [statusWebhook, setStatusWebhook] = useState<WebHook>(DEFAULT_WEBHOOK);
+  const [initialStatusWebhook] = useState<boolean>(false);
+
+  const refMessageWebhookUser = useRef<HTMLInputElement>(null);
+  const refMessageWebhookPass = useRef<HTMLInputElement>(null);
+  const [messageWebhook, setMessageWebhook] =
+    useState<WebHook>(DEFAULT_WEBHOOK);
+  const [initialMessageWebhook] = useState<boolean>(false);
```
kitajchuk, line 70: Looks like these initial state checks aren't working with the checkzones because you're not setting them in the `useEffect`. You can see how I did this on the accounts form, but basically for each one you check for the webhook username and password and either set true or false depending.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+                          .flat() as Voice[]
+                      }
+                      onChange={(e) => setSynthVoice(e.target.value)}
+                    />
+                  </fieldset>
+                </>
+              )}
+            </>
+          )}
+          <P>
```
kitajchuk, line 514: This example dummy code just needs to be removed now 👍

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+          `${webhook.label} username and password must be either both filled out or both empty.`
+        );
+        !webhook.stateVal.username && webhook.refUser.current?.focus();
+        !webhook.stateVal.password && webhook.refPass.current?.focus();
+        return;
+      }
+    });
+
+    if (applications) {
+      // update can still have the same name
```
kitajchuk, line 172: Again lets just keep the unnecessary comments out please 👍 . I'm not saying we don't use comments when it feels applicable, but in cases like this the code is quite clear enough to understand what is going on.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,595 @@
...
+  const [recogLang, setRecogLang] = useState<string>(LANG_EN_US);
+
+  const [synthesis, setSynthesis] = useState<SynthesisVendors | null>(null);
+  const [recognizers, setRecognizers] = useState<RecognizerVendors | null>(
+    null
+  );
+
+  const [message, setMessage] = useState("");
+
+  const webhook_method = [
```
kitajchuk, line 88: This is cool. I riffed on this in a separate PR because it's a concept we may want reused in other forms so once this PR is merged and the other one as well we can take a look at how to pull it in here.

You can see here: https://github.com/jambonz/jambonz-webapp/pull/80/files#diff-68572da928b5651e3f8dda9d2b291815dc0cdf0e0ff5dc40ab3dc81215b760fe

kitajchuk (changes requested):
Looking good but I have lots of feedback here for you to go through.

kitajchuk:
@khanghugo I also just added a new little `html` structure so we can compact the webhook fieldsets a bit. Once this PR is merged we can go back in and make the change on this form as well 👍 . Looks like this:

Screen Shot 2022-08-15 at 9 08 13 AM

Commit on my PR branch is here: https://github.com/jambonz/jambonz-webapp/pull/80/commits/0de64ac23983cc7bd19474c3076deb1b6eb3427b

khanghugo:
Thanks for the feedback. Let me rebase this first then see.

kitajchuk:
> Thanks for the feedback. Let me rebase this first then see.

There's nothing to rebase since the base `refresh` branch hasn't been updated yet. I just updated that screenshot though because I made some more UI changes to the forms for grouping of fieldset elements. I think it's more clear now.

src/containers/internal/views/applications/form.tsx
kitajchuk: Ignore this feedback for now since it will break the generic `handleSetHook` function which is relying on taking the input elements `name` attribute to apply the value. I'll think of a different way to handle this... 🤔

kitajchuk: I've updated the accounts form to use the `webhook.prefix` so the field names are unique. This breaks the `handleSetHook` function as I mentioned so I've also updated that to pass the `name/value` in explicitly rather than use the more brittle method I had in there before. That looks like this:

```tsx
const handleSetHook = (
  name: string,
  value: string,
  hook: WebHook,
  setHook: (h: WebHook) => void
) => {
  if (hook) {
    setHook({
      ...hook,
      [name]: value,
    });
  }
};
```

And then the calls to it now look like this:

```tsx
handleSetHook(
  "url",
  e.target.value,
  webhook.stateVal,
  webhook.stateSet
)
```

kitajchuk: Once you make this change for the field names the `handleSetHook` hook method will break. I've simplified in the accounts form to just make direct calls in my onChange handler and remove the `handleSetHook` abstraction entirely:

```tsx
webhook.stateSet({
  ...webhook.stateVal,
  url: e.target.value,
});
```

kitajchuk: If you're interested in seeing what it looks like you can check out the commit on my PR branch here:

https://github.com/jambonz/jambonz-webapp/pull/80/commits/eae2daf402769755b25625c165b5821bd2f6ddb3

src/containers/internal/views/applications/form.tsx
khanghugo: The current form do have an empty option exactly like this. I think it is best to have it than a default option of a random account.

khanghugo: Welp, i am wrong.

khanghugo: Wait, I am correct. 

Just found out something really goofy. 

This is application form with one account.
image


This is application form with more than one account

image


I think it is still best to have the empty option to remind. But this really makes half page working with `required` and another with some validation. Let me think what is better.

src/containers/internal/views/applications/form.tsx
khanghugo: You did this, not me.

kitajchuk: Ha 😅

src/containers/internal/views/applications/form.tsx
kitajchuk: The `useApiData` method returns what the actual api returned. It would have returned an array when you fetch all applications but when fetching by ID it should return just a single object.

kitajchuk: Looks like you're right, but I would say this is wrong of the API to respond like this. I just checked the old code and they were using a `filter` to get the "current" application: https://github.com/jambonz/jambonz-webapp/blob/main/src/components/forms/ApplicationForm.js#L160-L163.

But, again, this is wrong of the API as the swagger docs even show it returns a single application.

Screen Shot 2022-08-15 at 1 35 16 PM

kitajchuk: @davehorton What do you make of this? The API is returning an `array` with a single application when fetching via `application_sid` as in `/Applications/:application_sid`. This appears incorrect as the swagger docs express you would receive a single data object and I would expect to receive a single data object much like when fetching an account by sid 🤔

kitajchuk: @khanghugo For now keep things as you have them but feel free to change the `type` to be a single `Application` as that is the expectation here. I feel like we need the API to be updated as this seems like a backend bug...

kitajchuk:
@khanghugo Another thing we will need for parity is the account filter selector on the applications list view. I'm going to make this out of scope for this PR and I'll add an issue specifically for this to be tackled after this add/edit flow is complete 👍 

Screen Shot 2022-08-15 at 1 48 09 PM

kitajchuk:
@khanghugo I merged my PR branch. Whenever you end up rebasing just take all your changes on the conflicted applications index file and you should be good 👍

src/containers/internal/views/applications/delete.tsx
```diff
@@ -0,0 +1,92 @@
...
+              ]) ||
+              item[key1 as keyof typeof item])) ||
+            "[None]"
+        }
+      </li>
+    </ul>
+  );
+};
+
+export const DeleteApplication = ({
```
kitajchuk, line 46: Hey this implementation is not quite done yet. I'm looking at the requirements and you'll need to check that the application in question is not currently associated with an Account or an MS Teams Tenant before deleting. So, this component is going to end up needing to look a lot more like the Account delete but with just the two data checks rather than the four used in that scenario.

kitajchuk (changes requested):

kitajchuk:
@khanghugo The reusable account filter looks like this. I have a PR open for it so when you're ready we can merge that and then you can rebase again and slot it into your work here. The implementation is straight forward so hopefully it makes sense when you add it. If not, just ping me and I can help 👍 

https://github.com/jambonz/jambonz-webapp/pull/82

Screen Shot 2022-08-15 at 3 29 29 PM

src/containers/internal/views/applications/index.tsx
```diff
@@ -1,260 +1,115 @@
...
-          </P>
-          <P>
-            Selected synthesis language:{" "}
-            <strong>{langSynth || "undefined"}</strong>.
-          </P>
-          <P>
-            Selected synthesis voice: <strong>{voice || "undefined"}</strong>.
-          </P>
-          {synthesis && (
+  const [applications, refetch] = useApiData<Application[]>("Applications");
```
kitajchuk, line 14: Looking at the current implementation this is not correct. Applications are fetched by account SID so you’re definitely going to need the account filter component I have in PR currently. It will be used for several views including this one, however it look a like the way the data is fetched/filtered varies. In this case as I mentioned you use a different API: `Accounts/:account_sid/Applications`. So in this case we will end up swapping the useApiData hook for a combination of local state plus getFetch. I know you have a lot of other feedback on your plate so focus on that and when you’re ready we can work out the details of these changes 🤘

kitajchuk (changes requested):

src/containers/internal/views/applications/index.tsx
khanghugo: I think I missed this one. There should have been an option to select the account to show up its own applications. Let me do that so this one would appropriately be here. I need to take a more good look at the original Application page and do things more properly. 

I also have some ideas in mind but let me finish implementing it and see.

kitajchuk: Please hold off on this piece and focus on the rest first. As I mentioned I have a generic helper component for this with styling so I want to make sure we sync and get that implemented.

kitajchuk: Well, I went ahead and merged the `AccountFilter` so if you wanna give it a shot you can. There's an example on the Speech index stub currently of rendering it. You would still need to change how you fetch the application data like I was saying and use the `getFetch` api method. The way I would do it is create a state var for applications and accountSid (passed to AccountFilter). Then add a useEffect that takes accountSid as a dependency and calls getFetch and sets the applications state with the result of those calls 🤔 . Something like that lol

khanghugo: Okay I got this one exactly like wanted for the index page, let me see what I can do to do correctly for edit page.

kitajchuk:
@khanghugo I have a working list of feedback for the latest on this branch but I will wait until you request another review. But I did want to pose one change in particular now regarding the application list index and how to handle the state between fetch and delete. Check this diff here which has one commit of changes I made on top of the latest from this branch you are working on here. This is the approach I'd like to see here:

https://github.com/jambonz/jambonz-webapp/compare/refresh-application...refresh-application-index-example

kitajchuk:
@khanghugo Your last commit is off base. We don't want any of those changes to the account-filter component—it should not use the selector component. Do you have time to catch up on a chat?

khanghugo:
Sure, I can revert that change on Account Filter.

Where do you want to start a chat? I have Slack. The email is ninanino19913344@gmail.com

kitajchuk:
> ninanino19913344@gmail.com

I'll send a google meet if that works :)

kitajchuk:
> > [ninanino19913344@gmail.com](mailto:ninanino19913344@gmail.com)
> 
> I'll send a google meet if that works :)

Invite sent hop on duder :)

khanghugo:
It says Getting ready for the past 2 minutes already. Let me try again.

khanghugo:
This commit has all the requested changes. Let me probe around a bit and see if there is anything wrong.

kitajchuk:
> This commit has all the requested changes. Let me probe around a bit and see if there is anything wrong.

Cool I'll take a look. I have another idea brewing I'll run by you in a different thread/PR or something as well :)

src/containers/internal/views/applications/delete.tsx
```diff
@@ -0,0 +1,132 @@
...
+            return acc + used[key as keyof InUse].length;
+          }, 0) === 0;
+
+        if (deletable) {
+          setIsDeletable(deletable);
+        } else {
+          setInUse(used);
+        }
+      }
+    });
```
kitajchuk, line 86: You need empty deps here like `, [])` because as it stands now this `useEffect` is recursively firing and crashing the app backend with too many requests.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,604 @@
...
+          toastSuccess("Application updated successfully");
+        })
+        .catch((error) => {
+          toastError(error.msg);
+        });
+    } else {
+      postApplication(payload)
+        .then(() => {
+          toastSuccess("Application created successfully");
+          navigate(`${ROUTE_INTERNAL_APPLICATIONS}`);
```
kitajchuk, line 188: Let's route the user directly into the application edit view from here as this feels like a better user experience and this is how I've set up the accounts add/edit form to work. So it would  look like this:

```ts
postApplication(payload)
  .then(({ json }) => {
    toastSuccess("Application created successfully");
    navigate(`${ROUTE_INTERNAL_APPLICATIONS}/${json.sid}/edit`);
  })
```

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,604 @@
...
+        .catch((error) => {
+          toastError(error.msg);
+        });
+    }
+  };
+
+  useEffect(() => {
+    let ignore = false;
+
+    if (accounts && !accountSid) setAccountSid(accounts[0].account_sid);
```
kitajchuk, line 199: I'm not a big fan of the no brackets syntax used here for all of these, however I understand that is a personal nit lol. I guess maybe you like it then, or even prefer it?

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,604 @@
...
+                    }))}
+                    onChange={(e) => {
+                      setRecogLang(e.target.value);
+                    }}
+                  />
+                </fieldset>
+              )}
+            </>
+          )}
+          <fieldset>{message && <Message message={message} />}</fieldset>
```
kitajchuk, line 582: In order to avoid having an empty padded section when there is no message to render you'll have to change this to look like:

```tsx
{message && (
  <fieldset>
   <Message message={message} />
  </fieldset>
)}
```

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,604 @@
...
+      required: false,
+    },
+  ];
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+
+    setMessage("");
+
+    webhooks.map((webhook) => {
```
kitajchuk, line 131: I think the loop is a cool way to do this since I did each one individually on the account form (there's two) but this doesn't actually work. Returning from the `map` doesn't stop execution in this function block so the message renders but the form still submits. Playing around with it something like this works. This just stops on the first webhook error and shows the message.

```ts
const invalidHook = webhooks.find((webhook) => {
  return (
    (webhook.stateVal.username && !webhook.stateVal.password) ||
    (!webhook.stateVal.username && webhook.stateVal.password)
  );
});

if (invalidHook) {
  setMessage(
    `${invalidHook.label} webhook username and password must be either both filled out or both empty.`
  );
  !invalidHook.stateVal.username && invalidHook.refUser.current?.focus();
  !invalidHook.stateVal.password && invalidHook.refPass.current?.focus();
  return;
}
```

kitajchuk: I'm beginning to think we want to render errors next to the fields we focus because this jumps you back up the form to focus the field and you don't see the message at the bottom. I have maybe 3 ideas about how we could rethink showing error messages like this:

1. Figure out a clean way to render messages alongside their field of focus—this would need to be something nice and abstract so it's not verbose so I think we'd have to leverage the refs and break out of the traditional React rendering format. Not sure yet. Like, I don't want a state value for each possible field to check and render a message in each spot you know 🤔.
2. We simply just leverage the toast so that it always renders at the top—this is not a bad idea. We would have to modify the toast to have an option to render without a timeout, or a longer timeout, and include a close button so the user can disregard it once they consume the message.
3. We implement conditional `required` attributes on fields that, heh, "require" it. This would be nice because we'd be leveraging the native browser feature set here but we'd end up having variety logic on field element props to either render as required or not.

Which of these do you like? I'm generally leaning towards the toast option, 2, because it would be the most direct translation from what we have now.

This isn't something to implement on this PR so don't worry about it. I think I would spin off a branch to test this out and we can see what works best there. But it's something I'm thinking about now.

kitajchuk (changes requested):
Just about there dude! Great work! A few things noted here that should wrap this up :)

src/containers/internal/views/applications/delete.tsx
khanghugo: Wow. This is why. I was investigating this. Haha

kitajchuk: Yes—this is a weird thing about `useEffect` but if you use an empty dependency array it basically means "only run this once". Although in development React does the double dip so maybe you've noticed that everything renders twice because React mounts, unmounts and then mounts again. This is NOT the behavior in production mind you lol. So even in development you'll see these empty dep effects fire twice but reset assure that is a one time thing upon mount of the component.

src/containers/internal/views/applications/form.tsx
khanghugo: Yea I can have it with brackets for consistency sake.

src/containers/internal/views/applications/delete.tsx
khanghugo: That clears me up too. I was confused why it fetches so many times when the page only load once. I thought it could have been another oopsie somewhere. But glad to know that it is for development stage. I think I do really need to read some React related blog posts because I am right now just learning new things as I go.

kitajchuk:
Hey looks like that api server update was merged. I think Dave mentioned to
hold off on that until we're ready with the webapp refresh so I'm not sure
what we want to do from here? I suppose I should not have approved the PR
so I might have given the wrong implication there.

—BK

*   Brandon Lee Kitajchuk*

    *kitajchuk.com* <https://kitajchuk.com>

khanghugo:
Sorry for that. I have made a PR to revert the change.

src/containers/internal/views/applications/delete.tsx
kitajchuk: You should for sure. It's easy to make mistakes with React that can have disastrous consequences—especially with the whole hooks paradigm. In all honesty, I'm not a big fan of React or the ecosystem overall. However, it's clearly very popular and widely used so it makes sense for these projects. Here's a few good resources you could spend time reading:

https://beta.reactjs.org/learn/
https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/

The current react docs are good too for the general library API overview. But these newer beta docs have some great insights regarding real world implementation of React:

https://beta.reactjs.org/learn/you-might-not-need-an-effect#fetching-data

kitajchuk:
Heh, we'll see what Dave says about it. Easy enough to revert I suppose
even if it's an admin force push to main to reset the timeline—which this
might be preferential if we don't want the extra cruft in the timeline of
the main branch.

—BK

*   Brandon Lee Kitajchuk*

    *kitajchuk.com* <https://kitajchuk.com>

src/containers/internal/views/applications/form.tsx
kitajchuk: @khanghugo there's a curious bug when the webhook error messages are triggered. When I clear out one of either username or password and save the form I get the error message and the empty field focuses, however it gets the value from the data reapplied so it is no longer empty lol. Looks like an effect issue so I'm looking into that because I think the extra dependencies here are the culprit. I'll have recommendations here soon on how to fix this :)

khanghugo: I am down for consistency and option 3 seems just what we want. I really want to go with 2 but I don't really understand what you mean. Would that be a toast (head up block) rendering at the top of the page (maybe the screen) that can be dismissed? Would that sort of change how the `message` is rendered? I think we can try that and see.

kitajchuk: looks like that bug is due to using the toast for messages about form field errors. This is because the toast is in global state so it causes a waterfall render on the application. Not ideal for this scenario. Makes me consider rethinking storing the toast in this way, but I'll have to poke around that later.

kitajchuk: I do have some feedback now about separating out your effects a bit to decouple concerns. I'll post a diff so you can see what I mean.

kitajchuk:
@khanghugo Here's a diff that shows decoupling the effects. I did some other stuff like moving the fetches for applications and accounts up to the edit component. I've been preferring this approach (I did this with the account edit form as well to pass applications down) but honestly it's not necessary and maybe just a preference.

https://github.com/jambonz/jambonz-webapp/compare/refresh-application...refresh-application-effects-decouple

khanghugo:
That is a nice addition. The `add` page will now look less empty and `form` will do less stuffs.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,611 @@
...
+} from "src/api/types";
+
+export type UseApplicationData = {
+  data: Application | null;
+  error: FetchError | null;
+  refetch: () => void;
+};
+
+type ApplicationFormProps = {
+  accounts?: null | Account[];
```
kitajchuk, line 43: Cool new thing: since we’re passing these props from the add/edit parent components we can make the `accounts` and `applications` props non-optional since they are technically now required 🙂

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,611 @@
...
+                    onChange={(e) => {
+                      setRecogLang(e.target.value);
+                    }}
+                  />
+                </fieldset>
+              )}
+            </>
+          )}
+          {message && (
+            <fieldset>{message && <Message message={message} />}</fieldset>
```
kitajchuk, line 588: Lol you don’t need the extra `message && …` here inside the field set since the whole thing is wrapped in that condition.

src/containers/internal/views/applications/form.tsx
kitajchuk: > I am down for consistency and option 3 seems just what we want. I really want to go with 2 but I don't really understand what you mean. Would that be a toast (head up block) rendering at the top of the page (maybe the screen) that can be dismissed? Would that sort of change how the `message` is rendered? I think we can try that and see.

So after thinking on it a bit I agree with you that option 3 is the way to go for sure. Let’s do it. I’m going to make this change on the account form as well so we can compare implementation but I’m guessing we’re going to end up with roughly the same pattern 🤔🙂. We will still have the message for things like the duplicate name checks so I think we still can think about how to cleanly associate a message with a field that’s not too cumbersome. But we can worry about that later. The carrier form will demand a stronger need for it anyway.

khanghugo: I have looked around to see the implementations of custom messages for `input` tag and it seems like we can make another `form` component out of that.

kitajchuk: Cool. Drop some links. I have ideas too but I still think the conditional required model is nicer for the scenarios in which it can be applied.

khanghugo: Ehh, it is just "html input required message" and I click on the top result. But yea, conditional required is better where it needs, like duplicated application name. But for the webhook thing, `required` with some logic could be consistent. As if `required` is on the user side while conditional required is on the server side.

kitajchuk: Ya. I'm thinking maybe we lean into a more traditional model of usability regarding this stuff and just have clear labeling etc and also do the conditional required attribute stuff for things like the webhooks basic auth. Some thoughts that come to mind:

1. Use a UI indicator that a field is required, e.g. an asterisk with the label in our pink color
2. Have a top-level form info text that says something like "Fields marked with * are required"
3. Have field-level info text for nuanced required things like webhooks, e.g. "When using basic auth both the username and password fields are required"

If we were to follow this pattern the forms become more clear at the outset and we don't need to worry that the browser default messaging for the `required` attribute is non-descriptive. What do you think about these ideas?

kitajchuk: @khanghugo This worked out really great for the settings and accounts forms. Here's the commit on the `refresh` branch: https://github.com/jambonz/jambonz-webapp/pull/64/commits/42dff877b6b9c59b7b3d848c34e356f8f3e073e5

Here's a couple screenshots showing the implementations:

Screen Shot 2022-08-18 at 10 33 51 AM
Screen Shot 2022-08-18 at 10 34 04 AM

khanghugo: Okay. I am not sure why we have not done it before. It looks great. This might be really what we really want. I will have that for the application page too.

khanghugo:
I really like this change. It makes it clear what is expected rather than spitting out only if things happen. Also, the page looks less empty.

src/containers/internal/views/applications/form.tsx
kitajchuk: Ya it feels much better than what we had I agree.

kitajchuk:
> I really like this change. It makes it clear what is expected rather than spitting out only if things happen. Also, the page looks less empty.

Killer! If you’re ready I’ll give this a final review tomorrow and hopefully we’re ready to merge after that 🙂🤘

khanghugo:
It should be ready. I don't think I can really eye the small details that good since I am the one making it and thinking it is okay to go.

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,608 @@
...
+      initialCheck: initialStatusWebhook,
+      required: true,
+    },
+    {
+      label: "Messaging",
+      prefix: "message_webhook",
+      stateVal: messageWebhook,
+      stateSet: setMessageWebhook,
+      initialCheck: initialMessageWebhook,
+      required: false,
```
kitajchuk, line 107: This is nice 👍

src/containers/internal/views/applications/form.tsx
```diff
@@ -0,0 +1,608 @@
...
+                    placeholder="Optional"
+                    value={webhook.stateVal?.username || ""}
+                    onChange={(e) => {
+                      webhook.stateSet({
+                        ...webhook.stateVal,
+                        username: e.target.value,
+                      });
+                    }}
+                    required={
+                      webhook.required &&
```
kitajchuk, line 397: Nice way to do this :)

kitajchuk (approved):
This looks great dude—good work! Thanks for rallying all the back-and-forth with me. I feel like we resolved some UI/UX stuff to a really nice place now. Moving forward we should be looking solid.

Go ahead and smash that merge button and please make sure you use the `Squash and merge` option (which I think is the default now) so that we only get a single commit back to the base `refresh` branch! I would recommend the Speech form next :)

kitajchuk:
@khanghugo I'm going to go ahead and merge this one because I made some changes locally on the `refresh` branch that will require adding a couple classNames to the form here. This way I can merge, rebase and apply then push all latest up to the base `refresh` branch 👍

kitajchuk:
> @khanghugo I'm going to go ahead and merge this one because I made some changes locally on the `refresh` branch that will require adding a couple classNames to the form here. This way I can merge, rebase and apply then push all latest up to the base `refresh` branch 👍

@khanghugo Alrighty all merged and rebased so all the latest and greatest is on the base `refresh` branch for when you're ready to tackle your next issue 👍

khanghugo:
> Thanks for rallying all the back-and-forth with me

I am hoping that you are the one that doesn't mind the back and forth. I am all in for suggestions but it always come from you and that is what I hope you won't mind.


> I would recommend the Speech form next :)

Noted

> I'm going to go ahead and merge this one 

Yea, I wasn't online earlier to do this. It is alright. I will catch up and smash it myself next time.

kitajchuk:
> > Thanks for rallying all the back-and-forth with me
> 
> I am hoping that you are the one that doesn't mind the back and forth. I am all in for suggestions but it always come from you and that is what I hope you won't mind.
> 
> > I would recommend the Speech form next :)
> 
> Noted
> 
> > I'm going to go ahead and merge this one
> 
> Yea, I wasn't online earlier to do this. It is alright. I will catch up and smash it myself next time.

Ha, I don't mind at all. I enjoy the collaboration and as we can see we get the results: we ended up with some great improvements to how we handle forms for the app which is great :)

Issue #73, opened 2022-08-10
Refresh: Phone Numbers view


kitajchuk:
Implement functional parity for the `Phone Numbers` view form with the new UI style.

Create a branch from `refresh` called `refresh-phone-numbers` and submit a PR to the `refresh` branch. Please include this issue number (`#73`) in the commit/PR message.

You can find the container stub for this view [here](https://github.com/jambonz/jambonz-webapp/tree/refresh/src/containers/internal/views/phone-numbers). The list view also needs to include the multi-select + Add Application functionality.

New routes to be added:

- `/internal/phone-numbers/add`
- `/internal/phone-numbers/phone_number_sid/edit`

For any questions about design/UI/coding ping @kitajchuk.

Ultimately, when merging PRs, please always use "Squash and Merge" for the strategy.

PR #85, opened 2022-08-21
Phone number page for refresh

khanghugo:
#73

- [x] having all required pages
- [x] working routes
- [x] functional form
- [x] functional
- [x] integrating with voipcarrier (sort of stub because there is no voipcarrier type yet, but just a flip of a switch away)
- [x] navigate when account/carriers are not added
- [x] mass edit for used application in index page (seems like it will need more thing)
- [x] index page parity
- [x] styling for `disabled` of input 
- [x] stripping comments
- [x] full parity


khanghugo:
Will probably be done by tomorrow, easy peasy.

khanghugo:
It has all the features and work but the styling for `disabled` attribute for `input` tag is yet to look right. 

Besides that, there is this one feature from previous webapp, which is mass edit in the index page. This is what it looks like.
image

Also the index being a bit verbose with what is going on. That can be implemented easily but that will move some other get fetch functions from other views to something like `src/api/index` so it can be reused.

Some of the stuffs here will be fully implemented as soon as the type `VoipCarrier` is implemented. Right now, some of arbitrary variables are set to make it render.

kitajchuk:
> It has all the features and work but the styling for `disabled` attribute for `input` tag is yet to look right.
> 
> Besides that, there is this one feature from previous webapp, which is mass edit in the index page. This is what it looks like. image
> 
> Also the index being a bit verbose with what is going on. That can be implemented easily but that will move some other get fetch functions from other views to something like `src/api/index` so it can be reused.
> 
> Some of the stuffs here will be fully implemented as soon as the type `VoipCarrier` is implemented. Right now, some of arbitrary variables are set to make it render.

Ya. I’ll have to work with you on this because I need to come up with the styling for the new components, e.g. mass edit and disabled etc.

khanghugo:
Sure. I can add some stuff related to mass edit for index page meanwhile

khanghugo:
What do you think of having some info in the entry of each item on the index page? 

image

That would need to move some of the stuffs like `getApplications()` to the `src/api/types`.

khanghugo:
Sure sure. I will bounce on another issue then.

khanghugo:
Got that. I will stick around.

kitajchuk:
@khanghugo looks like you have conflicts. go ahead and resolve those and then lemme know what's blocking you on this one of if you just want me to do an initial review pass on it :)

khanghugo:
Oh.... I thought you were doing something with it so I have not touched it lately and been waiting for your commit to continue. Let me see what else it needs for parity. Meanwhile, you should check out the Ms Tenant view. It seems simple enough and already done.

kitajchuk:
> Oh.... I thought you were doing something with it so I have not touched it lately and been waiting for your commit to continue. Let me see what else it needs for parity. Meanwhile, you should check out the Ms Tenant view. It seems simple enough and already done.

Okay I think I’m remembering that convo now lol. Remind me what it was: mass edit UI correct? Maybe if you can get the functionality in there I can take a look at the new UI styling that needs to be added. If it’s ready for me to poke around and add styling then I can. Just lemme know 🙂

khanghugo:
> I can poke around and provide feedback. It looks like maybe you implemented mass editing but maybe just don’t have styling for it?

Let me make one quickly and see.

khanghugo:
> I’m also seeing that the carriers are a dependency here. That’s okay. You can run the old webapp on a different port and create some.

I guess then I will add the `VoipCarrier` type in this PR then.

khanghugo:
It is here, without styles, or really good UX. I cannot really get the checkbox to uncheck after mass edit.

There is an issue with mass edit.

The select-able applications could be unavailable in that account. So there is this case where the phone number is assigned with an account and an optional application that is not from that account. During edit, the webapp does not allow that behavior by filtering account_sid. But the index page mass edit is free for all. It does push changes to the back end. But it will be confusing if anyone tries to edit it. it can be confusing already if that is is able to change.

Besides that, this works.

kitajchuk:
> It is here, without styles, or really good UX. I cannot really get the checkbox to uncheck after mass edit.
> 
> There is an issue with mass edit.
> 
> The select-able applications could be unavailable in that account. So there is this case where the phone number is assigned with an account and an optional application that is not from that account. During edit, the webapp does not allow that behavior by filtering account_sid. But the index page mass edit is free for all. It does push changes to the back end. It will be confusing if anyone tries to edit it. it can be confusing already if that is is able to change.
> 
> Besides that, this works.

Does the current webapp also have this issue with the applications not mapping to the account?

khanghugo:
Right now, the behavior in this refresh is exactly what the webapp is. In the Add/Edit, when selecting an account, the optional application will change accordingly. When do mass edit, it will force the application regardless whether it belongs to an account or not. After that, when we edit the odd phone number, there is no application shown, due to fetch stuffs. But in the sql, it is there.

kitajchuk:
@khanghugo Okay I pushed my commit to refine the UI for mass edit and clean a few things up. Go ahead and test it out and see if it's all good on your end. UI now looks like this:

Screen Shot 2022-08-24 at 2 55 52 PM

kitajchuk (dismissed):

khanghugo:
Oh wow, so you have the stuffs in item info done already? It looks really nice nice. The ApplicationSelect being similar to AccountFilter is nice.

Looking forward to see what's see.

Issue #74, opened 2022-08-10
Refresh: MS Teams Tenants view


kitajchuk:
Implement functional parity for the `MS Teams Tenants` view form with the new UI style.

Create a branch from `refresh` called `refresh-ms-teams-tenants` and submit a PR to the `refresh` branch. Please include this issue number (`#74`) in the commit/PR message.

You can find the container stub for this view [here](https://github.com/jambonz/jambonz-webapp/tree/refresh/src/containers/internal/views/ms-teams-tenants). Review the `Accounts` view implementation [here](https://github.com/jambonz/jambonz-webapp/tree/refresh/src/containers/internal/views/accounts) to get an idea of the new approach to implementing views a la the `list/add/edit` model.

New routes to be added:

- `/internal/ms-teams-tenants/add`
- `/internal/ms-teams-tenants/ms_teams_tenant_sid/edit`

For any questions about design/UI/coding ping @kitajchuk.

Ultimately, when merging PRs, please always use "Squash and Merge" for the strategy.

PR #86, opened 2022-08-23
Microsoft Tenant page for refresh

khanghugo:
#74 

Just a quick copy-paste job. I still make sure that it works. Unless there are some places to poke.

### DIFF
--- a/src/containers/internal/views/ms-teams-tenants/add.tsx
+++ b/src/containers/internal/views/ms-teams-tenants/add.tsx
@@ -1,13 +1,13 @@
 import React from "react";
 import { H1 } from "jambonz-ui";
 
-import { useApiData, useServiceProviderData } from "src/api";
+import { useApiData } from "src/api";
 import { MsTeamsTenantForm } from "./form";
 
 import type { Account, Application, MSTeamsTenant } from "src/api/types";
 
 export const AddMsTeamsTenant = () => {
-  const [accounts] = useServiceProviderData<Account[]>("Accounts");
+  const [accounts] = useApiData<Account[]>("Accounts");
   const [applications] = useApiData<Application[]>("Applications");
   const [msTeamsTenants] = useApiData<MSTeamsTenant[]>("MicrosoftTeamsTenants");
 
--- a/src/containers/internal/views/ms-teams-tenants/add.tsx
+++ b/src/containers/internal/views/ms-teams-tenants/add.tsx
@@ -2,13 +2,11 @@ import React from "react";
 import { H1 } from "jambonz-ui";
 
 import { useApiData, useServiceProviderData } from "src/api";
-import { useSelectState } from "src/store";
 import { MsTeamsTenantForm } from "./form";
 
 import type { Account, Application, MSTeamsTenant } from "src/api/types";
 
 export const AddMsTeamsTenant = () => {
-  const currentServiceProvider = useSelectState("currentServiceProvider");
   const [accounts] = useServiceProviderData<Account[]>("Accounts");
   const [applications] = useApiData<Application[]>("Applications");
   const [msTeamsTenants] = useApiData<MSTeamsTenant[]>("MicrosoftTeamsTenants");
@@ -20,7 +18,6 @@ export const AddMsTeamsTenant = () => {
         accounts={accounts}
         applications={applications}
         msTeamsTenants={msTeamsTenants}
-        currentServiceProvider={currentServiceProvider}
       />
     </>
   );
--- a/src/containers/internal/views/ms-teams-tenants/add.tsx
+++ b/src/containers/internal/views/ms-teams-tenants/add.tsx
@@ -1,0 +1,29 @@
+import React from "react";
+import { H1 } from "jambonz-ui";
+
+import { useApiData, useServiceProviderData } from "src/api";
+import { useSelectState } from "src/store";
+import { MsTeamsTenantForm } from "./form";
+
+import type { Account, Application, MSTeamsTenant } from "src/api/types";
+
+export const AddMsTeamsTenant = () => {
+  const currentServiceProvider = useSelectState("currentServiceProvider");
+  const [accounts] = useServiceProviderData<Account[]>("Accounts");
+  const [applications] = useApiData<Application[]>("Applications");
+  const [msTeamsTenants] = useApiData<MSTeamsTenant[]>("MicrosoftTeamsTenants");
+
+  return (
+    <>
+      <H1>Add a Microsoft Teams Tenant</H1>
+      <MsTeamsTenantForm
+        accounts={accounts}
+        applications={applications}
+        msTeamsTenants={msTeamsTenants}
+        currentServiceProvider={currentServiceProvider}
+      />
+    </>
+  );
+};
+
+export default AddMsTeamsTenant;
--- a/src/containers/internal/views/ms-teams-tenants/form.tsx
+++ b/src/containers/internal/views/ms-teams-tenants/form.tsx
@@ -7,10 +7,7 @@ import { Section } from "src/components";
 import { Message, Selector } from "src/components/forms";
 import { MSG_REQUIRED_FIELDS } from "src/constants";
 import { toastError, toastSuccess, useSelectState } from "src/store";
-import {
-  ROUTE_INTERNAL_ACCOUNTS,
-  ROUTE_INTERNAL_MS_TEAMS_TENANTS,
-} from "src/router/routes";
+import { ROUTE_INTERNAL_MS_TEAMS_TENANTS } from "src/router/routes";
 
 import type {
   Account,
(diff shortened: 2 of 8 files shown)


khanghugo:
Just tested this and somehow it wasn't working for adding. Let me fix this.

kitajchuk:
@khanghugo I just rebased this one with the web button here so it’s all up to date. I’ll do a review while you work out what you can on the phone numbers PR 🤘

src/containers/internal/views/ms-teams-tenants/form.tsx
```diff
@@ -0,0 +1,202 @@
...
+  ServiceProvider,
+  UseApiDataMap,
+} from "src/api/types";
+
+type MsTeamsTenantFormProps = {
+  accounts: null | Account[];
+  applications: null | Application[];
+  msTeamsTenants: null | MSTeamsTenant[];
+  currentServiceProvider: null | ServiceProvider;
+  msTeamsTenant?: UseApiDataMap<MSTeamsTenant>;
```
kitajchuk, line 28: Ha! You got this, cool. I thought this was a nice generic refactor for this common pattern 🙂

src/containers/internal/views/ms-teams-tenants/form.tsx
```diff
@@ -0,0 +1,202 @@
...
+      if (msTeamsTenant.data.application_sid) {
+        setApplicationSid(msTeamsTenant.data.application_sid);
+      }
+    }
+  }, [msTeamsTenant]);
+
+  useEffect(() => {
+    if (accounts && !accounts.length) {
+      toastError(
+        "You must create an account before you can create an application."
```
kitajchuk, line 115: Maybe this is supposed to say “before you can create an MS Teams Tenant”?

src/containers/internal/views/ms-teams-tenants/form.tsx
```diff
@@ -0,0 +1,202 @@
...
+        {applications && (
+          <fieldset>
+            <label htmlFor="application_name">Application</label>
+            <Selector
+              id="application_name"
+              name="application_name"
+              value={applicationSid}
+              options={[
+                {
+                  name: "-- Optional --",
```
kitajchuk, line 170: I really don’t like this dash dash stuff so maybe we can make this just say “Choose application” and we can update the label to be “Application (optional)” if we feel it’s necessary to even say it. My thought is that with the note about required fields and the asterisks do we really need to spell out that this is optional?

src/containers/internal/views/ms-teams-tenants/add.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { H1 } from "jambonz-ui";
+
+import { useApiData, useServiceProviderData } from "src/api";
+import { useSelectState } from "src/store";
+import { MsTeamsTenantForm } from "./form";
+
+import type { Account, Application, MSTeamsTenant } from "src/api/types";
+
+export const AddMsTeamsTenant = () => {
+  const currentServiceProvider = useSelectState("currentServiceProvider");
```
kitajchuk, line 11: Probably move this state var to the form 🤔

src/containers/internal/views/ms-teams-tenants/form.tsx
khanghugo: Yea, maybe we don't need the "(optional)" in the label, just leave it there because there is no asterisks.

khanghugo:
Just saw this comment https://github.com/jambonz/jambonz-webapp/pull/64#issuecomment-1221096346.

Let me have that in these things right away.

kitajchuk:
> Just saw this comment [#64 (comment)](https://github.com/jambonz/jambonz-webapp/pull/64#issuecomment-1221096346).
> 
> Let me have that in these things right away.

Don’t worry about this I already fixed it when I found it lol

src/containers/internal/views/ms-teams-tenants/form.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const [message, setMessage] = useState("");
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+
+    setMessage("");
+
+    if (msTeamsTenants) {
+      if (
+        msTeamsTenants.find(
```
kitajchuk, line 53: Looking 👀 at this more closely now (the algorithm is the same on applications too) I’m thinking I prefer how we did this on the accounts submit here: https://github.com/jambonz/jambonz-webapp/blob/1f54bcd3c0f335832044a8d7f2dce024b81c78fb/src/containers/internal/views/accounts/form.tsx#L110. Basically, filter the list first then do the checks. I think this reads a lot better and helps the logic become more clear overall. What do you think?

src/containers/internal/views/ms-teams-tenants/form.tsx
```diff
@@ -0,0 +1,201 @@
...
+              a.ms_teams_tenant_sid !== msTeamsTenant.data.ms_teams_tenant_sid)
+        )
+      ) {
+        setMessage("The domain name you have entered is already in use.");
+        return;
+      }
+    }
+
+    const payload: Partial<MSTeamsTenant> = {
+      ...(!msTeamsTenant && {
```
kitajchuk, line 67: My preference would be to declare only the common fields here and spread the sid with the payload when passing it to the post method. I did this on the speech form submission. It reads more clearly and since we already have the if/else in place it feels preferable to do it that way.

src/containers/internal/views/ms-teams-tenants/form.tsx
```diff
@@ -0,0 +1,201 @@
...
+      if (msTeamsTenant.data.application_sid) {
+        setApplicationSid(msTeamsTenant.data.application_sid);
+      }
+    }
+  }, [msTeamsTenant]);
+
+  useEffect(() => {
+    if (accounts && !accounts.length) {
+      toastError(
+        "You must create an account before you can create an Microsoft Teams Tenant."
```
kitajchuk, line 114: I think if we’re going to say “Microsoft” instead of “MS” here we have to use “a” not “an” 🤔

src/containers/internal/views/ms-teams-tenants/form.tsx
khanghugo: Yea I can change to that. That also handles more conditions so it is definitely better for any future cases.

src/containers/internal/views/ms-teams-tenants/form.tsx
khanghugo: About this, should we move this check to the `add` page instead? Not only it is supposed to be adding but the page will look less empty. The form inherits some stuffs and add page is on higher level, I think it should handle that.

kitajchuk: Yes—look at how I implemented this on the Applications view. We need to leave this condition here as the fail safe, however you can greatly limit the opportunity for a user to end up here (probably by 100% lol) if you implement the model I did for Applications. It involves some clear messaging and not rendering the buttons for "add" if there are no applications. Maybe we can add the account filter to this list index as well? It's not there in the old webapp but since you have to connect the tenant to an account it would bring in parity of function between views 🤔 

Screen Shot 2022-08-24 at 8 33 14 AM

src/containers/internal/views/ms-teams-tenants/index.tsx
```diff
@@ -1,12 +1,132 @@
...
+        .catch((error) => {
+          toastError(error.msg);
+        });
+    }
+  };
+
+  return (
+    <>
+      <section className="mast">
+        <H1>Mircosoft Teams Tenants</H1>
```
kitajchuk, line 45: Typo here on "Microsoft" lol

src/containers/internal/views/ms-teams-tenants/form.tsx
khanghugo: I see. Let's just do the fail safe then. It makes more way more sense.

src/containers/internal/views/ms-teams-tenants/index.tsx
khanghugo: whooooooooops

kitajchuk:
@khanghugo I pushed a quick commit to fix an issue on this one regarding accounts. For parity this view is supposed to look at all accounts, not the specific service provider accounts. So that change resolves the need for the redirect even and now it functions the same as the old webapp. So as far as I can tell this one is done now 👍 

https://github.com/jambonz/jambonz-webapp/pull/86/commits/f1b117d10623fe6edd97eb24081b5701f2031f91

khanghugo:
Nice one. I will be standby for the next assignment then.

kitajchuk:
> @khanghugo I pushed a quick commit to fix an issue on this one regarding accounts. For parity this view is supposed to look at all accounts, not the specific service provider accounts. So that change resolves the need for the redirect even and now it functions the same as the old webapp. So as far as I can tell this one is done now 👍
> 
> [f1b117d](https://github.com/jambonz/jambonz-webapp/commit/f1b117d10623fe6edd97eb24081b5701f2031f91)

In looking into this more I'm not actually sure this makes sense. Maybe @davehorton can help clarify if this is working properly or not? We have exact parity with the old webapp on this branch now (with one key difference), so this is how it works and then maybe we can understand if this is correct or not.

**How this works currently:**

The MS Teams Tenants nav item is not displayed to the user unless the first service provider in the list of service providers fetched from the backend API has the `ms_teams_fqdn` field filled out. This turns out to be flaky ([code here for reference](https://github.com/jambonz/jambonz-webapp/blob/main/src/contexts/ShowMsTeamsContext.js#L15-L28)) in retrospect because I can have the `ms_teams_fqdn` for a service provider that is not zero in the Array index and never see the MS Teams Tenants nav item—which seems like a bug 🤔. The difference of how this works on the refresh webapp is we show/hide the MS Teams Tenants nav item based on whether the currently selected service provider has the `ms_teams_fqdn` field set.

**My question is:**

should the MS Teams Tenants be restricting the Accounts and Applications selectors to those of the current service provider only, or should they show all Accounts and Applications across all service providers which is how this feature currently operates.

Issue #71, opened 2022-08-10
Refresh: Carriers view


kitajchuk:
Implement functional parity for the `Carriers` view form with the new UI style.

Create a branch from `refresh` called `refresh-carriers` and submit a PR to the `refresh` branch. Please include this issue number (`#71`) in the commit/PR message.

You can find the container stub for this view [here](https://github.com/jambonz/jambonz-webapp/tree/refresh/src/containers/internal/views/carriers).

- [ ] The add/edit flow needs to have functional parity with current web app
- [ ] The list view also needs to include the accounts filter functionality

New routes to be added:

- `/internal/carriers/add`
- `/internal/carriers/voip_carrier_sid/edit`

For any questions about design/UI/coding ping @kitajchuk.

Ultimately, when merging PRs, please always use "Squash and Merge" for the strategy.

kitajchuk:
@khanghugo I've merged the MS Teams Tenants and the Phone Numbers PRs. They were looking/feeling good so we can be done with those. This form is up next. I'd like to work out the UI styling for tabs but I don't want that to hold you up on getting started since this is a big form. Go ahead and get started knowing that we're going to have to implement some new UI styling and semantics for the tabs, the gateways fields layouts and the IP whitelist layout. I don't think we need the Chip for SMPP gateways so we can just use a regular checkbox for that. So I'll let you get started and once you have the functionality in place lets connect and I can start getting some UI cleanup going on 👍

khanghugo:
Sure, I will have this one then.

kitajchuk:
@khanghugo I've decided to add the tabs to the `jambonz-ui` library so I have a PR up for that and there is a preview of the storybook deploy on Vercel:

PR: https://github.com/jambonz/jambonz-ui/pull/5
Preview: https://jambonz-ui-git-feature-tabs-jambonz.vercel.app/?path=/story/jambonz-ui-tabs--tabs-menu

This should be nice to use on the Carrier form 👍

khanghugo:
0.0.18 is not up on the registry. I guess I can do something else to have it.

kitajchuk:
> ~0.0.18 is not up on the registry. I guess I can do something else to have it.~ I thought it was merged. Whoops.

Heh, ya, not yet. I'm going to tweak a few more things and then get it merged and published.

kitajchuk:
[...]

PR #89, opened 2022-08-25
Carrier page for refresh 

khanghugo:
#71 

Remaining work to be done on this PR:

- [x] Proper SIP gateway validation (what we had before)
- [x] Proper SMPP gateway validation based on condition to validate
- [x] Implement refs for invalid gateway field highlighting
- [x] Properly validate port fields for empty so we don't get `Incorrect integer value: '' for column 'port' at row 1` (current web app has this issue)
- [x] Need to include Application selector if Account is set


src/api/index.ts
```diff
@@ -258,6 +260,10 @@ export const postPhoneNumber = (payload: Partial<PhoneNumber>) => {
   );
 };
 
+export const postCarrier = (payload: Partial<Carrier>) => {
+  return postFetch<SidResponse, Partial<PhoneNumber>>(API_CARRIERS, payload);
```
kitajchuk, line 264: I think this should be `Partial<Carrier>` lol

src/containers/internal/views/carriers/delete.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <Modal handleCancel={handleCancel} handleSubmit={handleSubmit}>
+      <P>
+        Are you sure you want to delete carrier <strong>{carrier.name}</strong>?
+      </P>
+    </Modal>
+  );
+};
+
+export default DeleteCarrier;
```
kitajchuk, line 28: I tend to avoid default exports unless necessary. In this case we can always use named exports and the only place we need default exports are for components tied directly to routes from the router. The reason those lazy routes do need default exports is because `React.lazy` operates on a promise using the dynamic `import` which requires a default export.

src/containers/internal/views/carriers/index.tsx
```diff
@@ -1,8 +1,122 @@
...
+import { Link } from "react-router-dom";
+import { Icons, Section, Spinner } from "src/components";
+import { hasLength } from "src/utils";
+import DeleteCarrier from "./delete";
 
 export const Carriers = () => {
-  return <H1>Carriers</H1>;
+  // the webapp fetches from all rather than jus the SP
+  // i guess we can fetch from SP but post in all?
+  const [carriers, refetch] = useServiceProviderData<Carrier[]>("VoipCarriers");
```
kitajchuk, line 15: The old webapp does not fetch from all, it's from the current SP:
https://github.com/jambonz/jambonz-webapp/blob/main/src/components/pages/internal/CarriersList.js#L111

Also we will post/put based on SP:
https://github.com/jambonz/jambonz-webapp/blob/main/src/components/forms/CarrierForm.js#L1006-L1012

khanghugo: Yea, not sure what I was thinking when writing this. Currently got some stuffs implemented/fixed accordingly. Would be done by the end of today, or maybe tomorrow morning.

src/api/index.ts
kitajchuk: Also this and the put method need to be by SP, so `/ServiceProviders/{sid}/VoipCarriers`

src/containers/internal/views/carriers/index.tsx
kitajchuk: No worries, but ya, looks like all the carriers stuff here is by SP ID rather than the loose endpoints. During this rewrite I've noticed some spottiness in the swagger docs so maybe we can get that tightened up at some point as well. I'll avoid too much commenting for now until you're really ready for it 👍

src/containers/internal/views/carriers/form.tsx
```diff
@@ -0,0 +1,492 @@
...
+                    id={`outbound_${i}`}
+                    name={`outbound_${i}`}
+                    type="checkbox"
+                    checked={g.outbound}
+                    onChange={(e) => {
+                      updateSipGateways(e, i, "outbound");
+                    }}
+                  />
+                  <div>Active</div>
+                  <button
```
kitajchuk, line 453: If you are going to put a button element in the form like this you have to add `type="button"` otherwise this button will submit the form, which it is currently doing.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -0,0 +1,492 @@
...
+              onChange={(e) => {
+                // TODO, clear the payload if these are unchecked
+                setPrefix(e.target.value);
+              }}
+            />
+          </Checkzone>
+        </fieldset>
+        <fieldset>
+          <Checkzone
+            // the checkbox size is scaling by how long the label is
```
kitajchuk, line 375: Ya this is a flexbox thing. It's easily addressed so don't worry about it for now :)

src/containers/internal/views/carriers/add.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { useApiData, useServiceProviderData } from "src/api";
+import { CarrierForm } from "./form";
+
+import type { Account, Carrier } from "src/api/types";
+
+export const AddCarrier = () => {
+  const [accounts] = useServiceProviderData<Account[]>("Accounts");
+  const [carriers] = useServiceProviderData<Carrier[]>("VoipCarriers");
+  const [predefinedCarriers] = useApiData<Carrier[]>("PredefinedCarriers");
```
kitajchuk, line 12: This is an interesting case. We may want to do some clever typing stuff to have a unique `PredefinedCarrier` interface since that has fields that don't actually exist on a `VoipCarrier` interface 🤔. You could rearrange things so that the `Carrier` interface is the common fields and then have `VoipCarrier` and `PredefinedCarrier` interfaces extend that and add their specific fields.

src/containers/internal/views/carriers/delete.tsx
khanghugo: :+1:

src/containers/internal/views/carriers/form.tsx
khanghugo: Yea, this is big. I per chance found out when I had to look up what the `type` of button means. Weird how `<button>` is not just a button. Haha.

kitajchuk: Semantically a button is meant to submit a form. The default value for its type attribute is “submit” for that reason. So the verbose type of “button” is to say: “you’re not meant to submit anything, but you’re an accessible element by default”.

khanghugo:
At this stage, most functionalities are here. The remainings are some more client side input validations. The webapp does some regex so that could be done easy. There are also some other conditions as well, like the application page. 

There this this one bug(?) where the `useEffect()` runs twice despite all the conditions but you have mentioned something with React development stage is doing things twice so that could be it. 

Besides that, things work and they all are here. Unless I miss some then I can come back and check again.

kitajchuk:
> At this stage, most functionalities are here. The remainings are some more client side input validations. The webapp does some regex so that could be done easy. There are also some other conditions as well, like the application page.
> 
> There this this one bug(?) where the `useEffect()` runs twice despite all the conditions but you have mentioned something with React development stage is doing things twice so that could be it.
> 
> Besides that, things work and they all are here. Unless I miss some then I can come back and check again.

Killer. I can do a review this weekend. Ya, effects will run twice in development with React. The idea is this should not break anything, and if it does you're likely doing something wrong lol 🤔.

davehorton:
I would love to see the latest, let me know which branch I should checkout and test...

kitajchuk:
> I would love to see the latest, let me know which branch I should checkout and test...

This branch is fine since it’s fully rebased with the base `refresh` branch. Note that this carrier form still needs UI work by myself and is not done yet (I’ve not yet even done a proper review of this PR as I was waiting so KL could get the functionality in there). You’ll need to run the backend on the branch for the obscure credentials PR to test out the speech form properly. But overall it’s feeling really nice! We should also hop on a chat/call to go over how to seed the dev db with data for recent calls and alerts since that’s what’s left next.

khanghugo:
All the stuffs are here. But I don't really like what I do for the latest commit. Pretty much all the check is copy paste from the old webapp with no closure. It seems pretty different from what we have done so far. Not sure how to get that implemented cleanly. There are some other minor stuffs that I have yet to figure out. But everything is here.

There is this one stuff that I don't add and that is adding a carrier from the predefined straight into the SP. In webapp, you can click from the preset and it will POST. But here, it will just occupy the form.

Everything is fine for me, just that the last commit is a bit too much.

kitajchuk:
> All the stuffs are here. But I don't really like what I do for the latest commit. Pretty much all the check is copy paste from the old webapp with no closure. It seems pretty different from what we have done so far. Not sure how to get that implemented cleanly. There are some other minor stuffs that I have yet to figure out. But everything is here.
> 
> There is this one stuff that I don't add and that is adding a carrier from the predefined straight into the SP. In webapp, you can click from the preset and it will POST. But here, it will just occupy the form.
> 
> Everything is fine for me, just that the last commit is a bit too much.

Cool no worries. I’ll review this weekend and can help with direction to get things cleaned up and in line with our new standards. I wanted to save this form as the last one for these reasons: it’s a beast lol. I have ideas so we may even introduce some new patterns here to make this one more maintainable. I also think we can break a lot of things out into utilities etc. I may do some UI work on this branch as well.

src/containers/internal/views/carriers/delete.tsx
```diff
@@ -0,0 +1,110 @@
...
+  handleCancel,
+  handleSubmit,
+}: DeleteProps) => {
+  const [inUse, setInUse] = useState<InUse | null>(null);
+  const [isDeletable, setIsDeletable] = useState(false);
+
+  useEffect(() => {
+    let ignore = false;
+
+    Promise.all([getFetch<PhoneNumber[]>(API_PHONE_NUMBERS)]).then(
```
kitajchuk, line 56: We don’t need a Promise.all for a single fetch. You can just fetch directly 👍

src/containers/internal/views/carriers/delete.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { useEffect, useState } from "react";
+import { P } from "jambonz-ui";
+
+import { Modal, ModalClose } from "src/components";
+
+import type { Carrier, PhoneNumber } from "src/api/types";
+import { getFetch } from "src/api";
+import { API_PHONE_NUMBERS } from "src/api/constants";
+import { hasLength } from "src/utils";
+// there is this error that is sent from the backend, not the frontend......
```
kitajchuk, line 10: Curious about this—I’ll have to run this branch and see but right now I’m doing a review on my phone lol

khanghugo: I think I forgot to remove that comment or something. I will take a look later.

src/containers/internal/views/carriers/delete.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const [inUse, setInUse] = useState<InUse | null>(null);
+  const [isDeletable, setIsDeletable] = useState(false);
+
+  useEffect(() => {
+    let ignore = false;
+
+    Promise.all([getFetch<PhoneNumber[]>(API_PHONE_NUMBERS)]).then(
+      ([phonesRes]) => {
+        if (!ignore) {
+          const used = {
```
kitajchuk, line 59: You can flatten all this logic since you only have one resource to fetch. You can get rid of most of it and just filter the response and check the length of that 🤔

src/containers/internal/views/carriers/delete.tsx
khanghugo: I think for this delete I have not implemented deletion for sip/smpp gateways also. That would be a quick addition. I have it in later when the review is done.

src/containers/internal/views/carriers/add.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const [carriers] = useServiceProviderData<Carrier[]>("VoipCarriers");
+  const [predefinedCarriers] =
+    useApiData<PredefinedCarriers[]>("PredefinedCarriers");
+
+  return (
+    <>
+      <H1>Add a speech service</H1>
+      <CarrierForm
+        accounts={accounts}
+        carriers={carriers} // currently unused
```
kitajchuk, line 20: Will it be used? If not then we don’t need it…

src/containers/internal/views/carriers/edit.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }, [error]);
+
+  return (
+    <>
+      <H1>Edit carrier</H1>
+      <CarrierForm
+        accounts={accounts}
+        carriers={carriers}
+        predefinedCarriers={predefinedCarriers}
+        carrierSipGateways={carrierSipGateways}
```
kitajchuk, line 55: You could normalize how you pass this data with its refetch methods to be more like how we pass the carrier interface here 🤔

kitajchuk: More like:

```tsx
sipGateways={{ data: sipGateways, refetch: sipRefetch }}
```

src/containers/internal/views/carriers/index.tsx
```diff
@@ -1,8 +1,154 @@
...
+  const [carrier, setCarrier] = useState<Carrier | null>(null);
+
+  // const [carriers, refetch] = useServiceProviderData<Carrier[]>("VoipCarriers");
+  const [carriers, setCarriers] = useState<Carrier[]>([]);
+  const [accounts] = useServiceProviderData<Account[]>("Accounts");
+  const [accountSid, setAccountSid] = useState("");
+
+  const [refetch, setRefetch] = useState(0);
+
+  const getCarriers = (all: boolean) => {
```
kitajchuk, line 25: I think it would be more preferable to use the `useServiceProviderData` and maybe just filter directly in the useEffect? Not sure because you’d still need a state var for the filtered carriers. Or you could filter on the fly in the render before you map but also not sure if that feels good or not 🤔.

src/containers/internal/views/carriers/index.tsx
```diff
@@ -1,8 +1,154 @@
...
+      </section>
+      <section className="filters filters--ender">
+        <AccountFilter
+          account={[accountSid, setAccountSid]}
+          accounts={accounts}
+          label="Used by"
+          defaultOption
+        />
+      </section>
+      <Section {...(hasLength(carriers) ? { slim: true } : {})}>
```
kitajchuk, line 80: Check out the other index pages. I tried to flatten the render logic a bit in those and would like to see that pattern here as well 👍

src/containers/internal/views/carriers/index.tsx
```diff
@@ -1,8 +1,154 @@
...
+        <div className="list">
+          {carriers ? (
+            hasLength(carriers) ? (
+              carriers.map((carrier) => (
```
kitajchuk, line 84: Ya I would just use the `useServiceProviderData` hook, skip the useEffect and filter inline here. It would simplify things and allow use of the refetch 🤔

src/containers/internal/views/carriers/form.tsx
```diff
@@ -0,0 +1,921 @@
...
+    }
+
+    const regIp =
+      /^((25[0-5]|2[0-4][0-9]|[0-1]?[0-9]?[0-9])\.){3}(25[0-5]|2[0-4][0-9]|[0-1]?[0-9]?[0-9])$/;
+    const regFqdn = /^([a-zA-Z0-9][^.]*)(\.[^.]+){2,}$/;
+    const regFqdnTopLevel = /^([a-zA-Z][^.]*)(\.[^.]+)$/;
+    const regPort = /^[0-9]+$/;
+
+    // copy pasted portion
+    // uhhh, not sure if this is what we want
```
kitajchuk, line 237: Probably not (at least not entirely or fully in this form) but let me do a proper review of the form when I can run this branch. Admittedly I’ve relaxed a bit this weekend and just spent time with my family so I’m just doing a quick pass here on my phone lol 😂

src/containers/internal/views/carriers/form.tsx
```diff
@@ -0,0 +1,921 @@
...
+              Save
+            </Button>
+          </ButtonGroup>
+        </fieldset>
+      </form>
+    </Section>
+  );
+};
+
+export default CarrierForm;
```
kitajchuk, line 921: You’ve reduced the size of this form by almost 1,000 lines of code so you should be proud of that 👏. We’ll get the questions sorted this week and dial it in!

src/containers/internal/views/carriers/delete.tsx
kitajchuk: That makes sense and in that case you’ll want this more robust logic 🙂

src/containers/internal/views/carriers/form.tsx
khanghugo: Thanks. Let me resolve these convos and implement the deletion of sip/smpp gateway when carrier is deleted.

src/containers/internal/views/carriers/index.tsx
khanghugo: I end up using a state to avoid duplication of filter. But it is possible just to filter twice here as I make the filter one line.

kitajchuk: Ya that's a fair point. I wasn't considering we need the filtered carriers passed to the `hasLength` function as well lol

kitajchuk:
@khanghugo I've pushed a commit with an initial round of UI and cleanup. Still plenty to do. I haven't resolved the SIP Gateway layout structure yet but I got started on it. That one is tough with so many fields for one grouping in terms of creating a nice responsive layout. The SMPP Gateways will be easy by comparison. I can do another UI pass on it tomorrow which should wrap things up in that regard and then maybe we can discuss some of the ideas regarding refactoring local utils etc to make this one even a bit cleaner. But overall it's feeling much nicer already :)

Here's the commit I pushed: https://github.com/jambonz/jambonz-webapp/pull/89/commits/5167fe37afc7356386640171521b55374446f361

khanghugo:
Nice work. Looks nicer already. I wasn't really sure what the difference between the `Smpp` and `Sbc` thing... But glad to see it was a quick change, not too far off.

I have come to accept that the regex part is fine. If it looks too out of place then we can just move it out of sight...

kitajchuk:
> Nice work. Looks nicer already. I wasn't really sure what the difference between the `Smpp` and `Sbc` thing... But glad to see it was a quick change, not too far off.
> 
> I have come to accept that the regex part is fine. If it looks too out of place then we can just move it out of sight...
I can poke at the regex stuff tomorrow as well when I do a final UI/cleanup pass on this one. I think we can make a move to put that stuff out of sight into some local utils because it’s an eye sore. Might be able to refactor as well so it’s better overall.

I also still am curious about the delete modal. If we’re not going to have more resources to check there we can just flatten that logic and UI to simplify 🤔.

khanghugo:
> I also still am curious about the delete modal. If we’re not going to have more resources to check there we can just flatten that logic and UI to simplify thinking.

Yea it does seem like it can be shorted way much more. I'd go with that because initially I just have those stuffs there to see if I have more than one stuff to check, the down to one stuff, then down to just only one stuff.

kitajchuk:
@khanghugo I've [pushed another commit](https://github.com/jambonz/jambonz-webapp/pull/89/commits/e864219a3070ad875209e4c9a6b308a1c096377b) with essentially final UI cleanup. It's suitable for sure so I think we can roll with it for now 👍. I've made a new checklist highlighting the remaining issues as I see it when comparing this PR to the old web app functionality. I think the remaining issues are pretty clear but we may have to switch our model here a bit—not sure. One primary issue with using the native form field `required` or `minlength` is that those fields are not focusable when hidden on the page and considered to be invalid by the browser. This breaks our current validation using this pattern. So we may need to resolve to using just the custom error messages here unless you see another way forward? If that's the case we will need the errors to be a local array in the handle submit function and we'd need to set them to the message state var using a map because that can accept a JSX tree which could be an error on each line. It could even be just a flat `ul`—I did this with the `MSG_PASSWD_CRITERIA` constant for password validation on login so you can see how that works. You would just need to make the message state var an IMessage: `const [message, setMessage] = useState<IMessage>("");`. In order to supplement wrapping validation in this way I will need to update the `jambonz-ui` tabs component to accept a prop for setting the active tab externally so we can change tabs programmatically during validation if need be. The current web app does this so it will likely be necessary to complete validation in this way. I'm hoping you can run with this direction and I can make the change on the tabs UI side and then we can put it together and wrap this one up 👍.

kitajchuk:
@khanghugo New version of `jambonz-ui` is `0.0.19` and this will have the update to the tabs component:

https://jambonz-ui.vercel.app/?path=/story/jambonz-ui-tabs--tabs-menu

khanghugo:
Alright. Let me see what I can do.

kitajchuk:
> Alright. Let me see what I can do.

Excellent. I was thinking we can probably do some smart logic to do a loop on the gateways by type (sip/smpp) and run the shared validation logic in that way to keep it DRY. Probably a couple ways you could achieve that so feel free to give it a shot.

Also I am making some other bothering changes here since it's the active PR branch and I'm too lazy to make new branches right now. I'm going to be pushing a commit that removes as much explicit `null` type definitions as possible as that was starting to bother me and I'd rather lean on `undefined` for cleanliness. Shouldn't get in your way, but be warned lol 🤞

khanghugo:
> One primary issue with using the native form field required or minlength is that those fields are not focusable when hidden on the page and considered to be invalid by the browser. This breaks our current validation using this pattern.

Yea, it does seem like using a message would be simpler. It can be easier to implement in comparison with having more states of HTMLElement. Doing an array of message isn't too bad, especially with this more complicated form. That is another quick addition also.

I am not sure about just doing a flat message as it will not be as helpful to point out the errors/requirement.

> I was thinking we can probably do some smart logic to do a loop on the gateways by type (sip/smpp) and run the shared validation logic in that way to keep it DRY.
That is doable. Let me try my best to make it not look too bad.

khanghugo:
> Disable option by static ip, requires changes to Selector tag (@khanghugo is this necessary?)

Eh, just my first thought. That could be scraped. I initially thought that because the webapp straight up POST the present once you select it. This carrier form just occupies the preset so I thought it would be a nice UX. Sure, it can be scraped because that would involve more stuffs, or a just quick hack of directing to `add` page, lol.

kitajchuk:
@khanghugo If you're not too far in do a fresh pull on this branch. Pushed some general tidying up and fixed a bug I introduced when using `filter` to check if we should map and render inbound/outbound smpp gateways 🙃

khanghugo:
@kitajchuk I have just been thinking a bit about the `required` stuffs. The fact that it is not focused because the element being hidden seems like a problem. I think we can do something like having the `required` condition coupled with active tab. This would enable the real validation we have from the `handleSubmit` and message being pump out regardless whether the element is hidden or not. So the `required` is to be sort of interactive validation while there is another validation after that. That would play nicely to what we already have without changing things too much. 

We don't use any HTMLElement stuff to focus, which would add on some more work, and I think it is better if we just keep it that what.

kitajchuk:
> @kitajchuk I have just been thinking a bit about the `required` stuffs. The fact that it is not focused because the element being hidden seems like a problem. I think we can do something like having the `required` condition coupled with active tab. This would enable the real validation we have from the `handleSubmit` and message being pump out regardless whether the element is hidden or not. So the `required` is to be sort of interactive validation while there is another validation after that. That would play nicely to what we already have without changing things too much.
> 
> We don't use any HTMLElement stuff to focus, which would add on some more work, and I think it is better if we just keep it that what.

Ya, I was just sort of resolving my headspace to a similar idea so I'm glad we're landing on the same page here. You may still end up with some tricky issues because the native required/minlength functionality I believe literally doesn't work when the inputs are hidden—so the logic regarding which tab to validate first and/or switch to might be the hard part to figure out 🤔

kitajchuk:
> > @kitajchuk I have just been thinking a bit about the `required` stuffs. The fact that it is not focused because the element being hidden seems like a problem. I think we can do something like having the `required` condition coupled with active tab. This would enable the real validation we have from the `handleSubmit` and message being pump out regardless whether the element is hidden or not. So the `required` is to be sort of interactive validation while there is another validation after that. That would play nicely to what we already have without changing things too much.
> > We don't use any HTMLElement stuff to focus, which would add on some more work, and I think it is better if we just keep it that what.
> 
> Ya, I was just sort of resolving my headspace to a similar idea so I'm glad we're landing on the same page here. You may still end up with some tricky issues—so the logic regarding which tab to validate first and/or switch to is the trick. I think it should be top to bottom (Voice => SMS). If you do it this way you can break on the first error each time and set the appropriate tab and hopefully the browser still handles the attributes for us 🤔. I'm not convinced this will work, however, since you would be setting the tab on submit and I presume the browser validation race condition will beat us on the tab switch.

@khanghugo I'm doing some quick testing and it looks like the form doesn't submit even though the required fields are hidden in the background tab so you should be able to find the right logic implementation to pull this off 👍

kitajchuk:
@khanghugo Setting the active tab in the form is not working for me so something is wrong there...

kitajchuk:
> @khanghugo Setting the active tab in the form is not working for me so something is wrong there...

Okay it works—I had to restart Vite after npm installing lol. That module cache is nice but oops 🙃. So you can perform some logic in an `onClick` on the submit button to determine which tab to switch to for proper form field validation. This worked for me and the tabs changed and the browser rendered the native required message 👍 

Screen Shot 2022-08-30 at 4 38 01 PM

khanghugo:
Another thought. In the webapp, sip gateway is required while smpp is optional. Maybe, we don't need tab at all and have the SMS/smpp an optional thing when enabled. When I first tried out the form, I am under the impression that the carrier can be either Voice or SMS or both. But it is just Voice, at least according to how the webapp works. So maybe we don't need to do these tabs and let the page go on a bit longer..

kitajchuk:
> 

I would keep it as is with the tabs for now since the separation is nice. I think the tab switching will be easy since you just need to leverage your existing state logic that already renders the required attributes. The SMPP fields do have validation but only for outbound. If you enter a system ID and password for outbound then you are required to enter at least one valid outbound SMPP gateway. Inbound is optional however the inbound gateway fields still use the same validation logic for ip/port etc.

khanghugo:
> Okay it works—I had to restart Vite after npm installing lol 

I need to know how to do this. I am getting this problem too.

khanghugo:
https://github.com/jambonz/jambonz-webapp/blob/2f4eb1cddafeaf3d3ac1bcccb5d515658694c77f/src/containers/internal/views/carriers/form.tsx#L939

Not sure how this plays out. The webapp does mention this part but it is commented out.
https://github.com/jambonz/jambonz-webapp/blob/4cfdfc3b49cf69ce3abf1996251709e84eac2d38/src/components/forms/CarrierForm.js#L868-L883

khanghugo:
> The old web app is potentially broken

I hope the parts I think is not broken is in fact not broken. This is getting confusing.

kitajchuk:
> > The old web app is potentially broken
> 
> I hope the parts I think is not broken is in fact not broken. This is getting confusing.

lol you're fine. up to this point we have things well in order.

however, i just tested the looped gateway logic and it doesn't seem to actually work. also i'm not sure i like the implementation at this point with the `for...in` model or the loop type variable. i have take a deeper look tomorrow and help out.

kitajchuk:
@khanghugo I pushed some more cleanup. Switched to the `number` input type for ports and that really resolves the need to even validate them anymore on submit (although I left it in for now…). Removed some unneeded checks on sip gateways due to our `required` fields etc. I have a short list of refinements (not on our checklist here) I’ll push tomorrow and then I think we just need to finish the couple checklist items 🙂.

khanghugo:
I see the way you do validation now. I have been avoiding using just a normal for loop because it looks different but I cannot really think of another solution that would make it look in place and work well (hence another variable keeping track of the enumeration of enhanced for loop). But if that seems fine then I guess we can also add another loop outside like the previously reverted commit. 

>  Switched to the number input type for ports and that really resolves the need to even validate them anymore on submit 

I initially did this but then I just make it into a select because it would look the same. I hope no one does that for inputting birthday.

I think this branch is now yours to caress? I don't want to just submit random commit while you are working.

kitajchuk:
@khanghugo Okay I've opened this one up for review. It feels ready to me but go ahead and play around with it and let me know what you think?

khanghugo:
The `handleActiveTab` seems to not change active tab only when the sip/smpp ip field is empty, per this condition.

https://github.com/jambonz/jambonz-webapp/blob/0615f7e500df13fbd4b4eaab691df40081aecf7d/src/containers/internal/views/carriers/form.tsx#L252

It could be a quick fix with adding setting active tab for every error in `handleSubmit` sip/smpp validation. But I tried that and it doesn't focus. Not sure what to add on for that.

kitajchuk:
> The `handleActiveTab` seems to change active tab only when the sip/smpp ip field is empty, per this condition.
> 
> https://github.com/jambonz/jambonz-webapp/blob/0615f7e500df13fbd4b4eaab691df40081aecf7d/src/containers/internal/views/carriers/form.tsx#L252
> 
> It could be a quick fix with adding setting active tab for every error in `handleSubmit` sip/smpp validation. But I tried that and it doesn't focus. Not sure what to add on for that.

It works in all the testing I've done. The rules are:

1. All SIP gateway fields must validate (are required) so empty IP in any row handles this. If you don't want the empty row, delete it and save the form
2. SMPP outbound gateways must validate ONLY IF you have entered either a system ID or password for outbound, otherwise they are not required and therefore it doesn't necessarily matter. Empty rows beyond at least one valid row don't hinder form submission and they are ignored. 
3. SMPP inbound gateways don't validate but the inbound password field is required if you have entered values for an inbound gateway
4. All SMPP gateway ip fields validate IF there is a value entered. Empty ip fields mean the row is ignored and form submission is not hindered.

All my testing has shown these business rules to operate correctly for this form. I am unable to replicate a scenario in which an empty SIP gateway IP doesn't change the tab and show the browser required message focusing the field 🤔

khanghugo:
Oh, I mean the tab only changes when the field is empty. I try to have some random wrong stuffs and it doesn't change tab.

kitajchuk:
> Oh, I mean the tab only changes when the field is empty. I try to have some random wrong stuffs and it doesn't change tab.

Ah, gotcha—yes that is true. Good callout. I can fix that pretty quick :)

kitajchuk:
> > Oh, I mean the tab only changes when the field is empty. I try to have some random wrong stuffs and it doesn't change tab.
> 
> Ah, gotcha—yes that is true. Good callout. I can fix that pretty quick :)

@khanghugo I pushed changes that fix that issue. Refactored validations into reusable getters to make this cleaner.

kitajchuk:
> > > Oh, I mean the tab only changes when the field is empty. I try to have some random wrong stuffs and it doesn't change tab.
> > 
> > 
> > Ah, gotcha—yes that is true. Good callout. I can fix that pretty quick :)
> 
> @khanghugo I pushed changes that fix that issue. Refactored validations into reusable getters to make this cleaner.

@khanghugo I just pushed one more commit that renders the gateway validation error message contextually near the fields that are invalid. I also pushed a commit that adds an IP regex `pattern` to the inbound gateway ip field since FQDNs are disallowed for this type of gateway. This basically makes the FQDN validation on this type unnecessary but I've left it in since we validate all SMPP in a single loop 🤔

kitajchuk:
@khanghugo Also just pushed sticky tabs lol. Maybe kind of nice, maybe just icing on the cake 🍰

khanghugo:
Wow, it looks really different how we let the browser handling some of the regex, rather than the script validation. I don't even know that those attributes exist.

kitajchuk:
> Wow, it looks really different how we let the browser handling some of the regex (and some other validations), rather than the script running. I don't even know that those attributes exist.

I think you can do a lot if you really embrace input types and constraint validations. Check out this page as some introductory reading on all this: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation

Issue #75, opened 2022-08-10
Refresh: Recent Calls view


kitajchuk:
Implement functional parity for the `Recent Calls` view form with the new UI style.

Create a branch from `refresh` called `refresh-recent-calls` and submit a PR to the `refresh` branch. Please include this issue number (`#75`) in the commit/PR message.

You can find the container stub for this view [here](https://github.com/jambonz/jambonz-webapp/tree/refresh/src/containers/internal/views/recent-calls). It already has some info and the logic stubbed for using mock data from a local dev server API implementation.

For any questions about design/UI/coding ping @kitajchuk.

Ultimately, when merging PRs, please always use "Squash and Merge" for the strategy.

khanghugo:
How would you like to proceed with these? At best I can make it have all the functionalities but I don't think I can just test them out.

kitajchuk:
> How would you like to proceed with these? At best I can make it have all the functionalities but I don't think I can just test them out.

Ideally we can work with @davehorton to get the dev db seeded with a robust enough data set for this and alerts so we can properly build and test all filters 🤔. Dave and I have spoken about this a bit already so it’s definitely on his radar. This is all we have left so we can prioritize how to go about this.

kitajchuk:
@khanghugo I finished a functional local dev server with responses for Recent Calls so you can get started using that. It's fully supportive for the UI and I've tested even with the old web app (replacing the axios URLs with our dev server ones) and it's fully functional for that UI 👍. You should be able to build the UI with feature parity using this so give it a shot and just reach out with questions or if you need help. We can probably do this with the Alerts view as well and this should be good enough for local dev with these views.

Commit: https://github.com/jambonz/jambonz-webapp/pull/64/commits/3012bbb7b3c7225a58c544325104c5aa02d074bc

How I changed the axios URLs in the old web app to verify full functionality of the dev server:

<img width="444" alt="Screen Shot 2022-09-01 at 4 51 00 PM" src="https://user-images.githubusercontent.com/438711/188033519-a9aa1c51-2cf6-4050-8cd4-e943a3c4e3ae.png">

As you can see it works just fine so it should be enough to build the new UI:

<img width="1494" alt="Screen Shot 2022-09-01 at 5 20 49 PM" src="https://user-images.githubusercontent.com/438711/188033615-710f3fd9-6fa7-48d4-9b6d-6cc15e36dc64.png">

PR #93, opened 2022-09-02
Recent call page for refresh

khanghugo:
#75 

**Checklist for UI completion:**

- [x] List table style (needs to be slimmed down etc)
- [x] Variant details/summary styles (needs to be clean and full width etc)
- [x] Pagination style (needs UI love and should use Icons etc)
- [x] Pcap only fetches when line item is toggled 🤔 
- [x] Status filter seems broken (sends page 0 which needs to never happen...)


src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -1,28 +1,99 @@
...
+    { name: "answered", value: "true" }, // string..........
+    { name: "not answered", value: "false" },
+  ];
+
+  const perPageSelection = [
+    { name: "25 / page", value: "25" },
+    { name: "50 / page", value: "50" },
+    { name: "100 / page", value: "100" },
+  ];
+
```
kitajchuk, line 68: This helper already exists in the utils so just import it from there 👍

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -32,12 +103,60 @@ export const RecentCalls = () => {
...
+  }, [
+    accountSid,
+    dateFilter,
+    directionFilter,
+    statusFilter,
+    perPageFilter,
+    pageNumber,
+  ]);
+
+  useEffect(() => {
```
kitajchuk, line 115: Curious about these effects 🤔. Feels like we can handle this logic more directly in the event handlers or fetch chains? Not sure, what’s your take?

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -32,12 +103,60 @@ export const RecentCalls = () => {
...
-      <Section>
+      <section className="mast">
+        <H1>Recent Calls</H1>
+      </section>
+      <section className="filters filters--ender">
+        <AccountFilter // half and half
+          account={[accountSid, setAccountSid]}
+          accounts={accounts}
+        />
+        <label htmlFor="date_filter">Date: </label>
```
kitajchuk, line 137: So this selector component is only for forms. The style of the account filter is the style we want for these as well. I’ve been considering a more generic version of that style as a reusable component that takes options because I made an application selector for the speech index that is also similar.

kitajchuk, line 137: So this selector component is only for forms. The style of the account filter is the style we want for these as well. I’ve been considering a more generic version of that style as a reusable component that takes options because I made an application selector for the speech index that is also similar.

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -92,7 +211,92 @@ export const RecentCalls = () => {
...
+                </div>
+              </div>
+            ))
+          ) : (
+            <M>No data</M>
+          )}
+        </div>
       </Section>
+
+      <ButtonGroup>
```
kitajchuk, line 245: I think the ideal way to do this would be to create a generic pagination component because we need it for the alerts view as well 🙂

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -92,7 +211,92 @@ export const RecentCalls = () => {
...
+            )}
+
+        <Button
+          disabled={pageNumber === maxPageNumber}
+          small
+          onClick={() => setPageNumber(pageNumber + 1)}
+        >
+          {">"}
+        </Button>
+        <Selector
```
kitajchuk, line 293: This also should have the smaller style and not use this particular component. Alternatively we could modify this component to have an optional prop that renders it in the smaller style 🤔

kitajchuk:
@khanghugo Glad you’re getting started! I’ll have a lot of UI/UX opinions for these views so as long as you get the functionality hooked up I can do a design audit and we can get things feeling flossy like the rest of the refresh app 🙂

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -1,28 +1,99 @@
...
+import type {
+  Account,
+  // Pcap,
+  RecentCall,
+} from "src/api/types";
+import { Selector } from "src/components/forms";
+import { hasLength, hasValue } from "src/utils";
 
 export const RecentCalls = () => {
-  const [pcap, setPcap] = useState<Pcap>();
```
kitajchuk, line 12: You will ultimately need the pcap stuff so best to keep it around 🤔

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -1,28 +1,99 @@
...
   useEffect(() => {
     let ignore = false;
 
     getRecentCalls("account-sid", {
-      page: 1,
-      count: 25,
-      start: dayjs().startOf("date").toISOString(),
+      ...(dateFilter === "today"
+        ? { start: dayjs().startOf("date").toISOString() }
+        : { days: parseInt(dateFilter) }),
```
kitajchuk, line 86: When using parseInt please always supply the radix, so 10 for us here.

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -92,7 +211,92 @@ export const RecentCalls = () => {
...
           </Button>
         )}
+      </Section> */}
+      <Section {...(hasLength(calls) ? { slim: true } : {})}>
+        <div className="list">
+          {!hasValue(calls) && <Spinner />}
+          {hasLength(calls) ? (
+            calls.map((call) => (
+              <div
+                className="item"
```
kitajchuk, line 221: This is fine for now but we will want to come up with a new markup and style for these views. Probably more table-like but I’m going to work out some UI once the functions are in place.

src/containers/internal/views/recent-calls/index.tsx
khanghugo: I wasn't aware pagination is a thing. I look at the original webapp and has no idea how it looks so tiny. I went so far to implement something like that inline and it is already a feature.

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -92,7 +211,92 @@ export const RecentCalls = () => {
...
+                key={`${call.call_sid}-${call.attempted_at}-p${pageNumber}`}
+              >
+                <div className="item__info">
+                  <div className="item__title">
+                    {/* i dont think this is working, it updates every second if button pressed TODO*/}
+                    {new Date(call.attempted_at).toLocaleString()}{" "}
```
kitajchuk, line 227: You should probably use dayjs here anyway for ease of formatting. Check out the npm or GitHub page for the api but in general it’s a lighter version of moment js.

src/containers/internal/views/recent-calls/index.tsx
kitajchuk: Not sure what you mean? We don’t yet have a pagination component for the refresh so my suggestion would be to make it and use it here rather that embedding it all in this isolated view.

src/containers/internal/views/recent-calls/index.tsx
khanghugo: Yea, I think we can do that like what the webapp does.

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -92,7 +211,92 @@ export const RecentCalls = () => {
...
+            <M>No data</M>
+          )}
+        </div>
       </Section>
+
+      <ButtonGroup>
+        <P>
+          Total: {callsTotal} record{callsTotal === 1 ? "" : "s"}
+        </P>
+        <Button
```
kitajchuk, line 249: So if by your tiny comment you refer to these buttons I would say don’t use this button component here. This isn’t really what this button style is meant for. I tend to use regular button elements with the `btnty` class and then style according to desired UI. Kind of like what I did with the plus and trash buttons for the gateways when I cleaned up the UI on the carrier form.

src/containers/internal/views/recent-calls/index.tsx
khanghugo: I think I understand this wrong. I thought pagination is a feature in HTML, but it seems like not.

kitajchuk: Ha! No it’s not. Ya my point is it’s a good opportunity to abstract the component so we can reuse it for the alerts view 🙂

src/containers/internal/views/recent-calls/index.tsx
khanghugo: How would you like to proceed with this? Do you want to make these changes after all the functionalities are done?

src/containers/internal/views/recent-calls/index.tsx
khanghugo: :+1:

src/containers/internal/views/recent-calls/index.tsx
khanghugo: Yea, I can have just the normal select then.

src/containers/internal/views/recent-calls/index.tsx
kitajchuk: For this in particular you could try making the abstract select component or modifying this one to render the alternate style. Overall I’m going to be doing a UI sweep but I’d like to wait until you’re done with the functionality—yes. I think we can do a lot of cool stuff here for UI/UX so it will be fun to show you what I mean when we get there 🙂

khanghugo: I have a generic defined in the index of recent call without any style to replace the parts that might need those later

src/containers/internal/views/recent-calls/index.tsx
khanghugo: I have a generic pagination but it looks exactly this but no style, defined in component, and trying a bit too much.

src/containers/internal/views/recent-calls/index.tsx
khanghugo: Hmmm, I cannot get the `className="btnty"` to work. It is just hidden. Probably has something to do with its being in a tag or something. But it is now there, nothing with it.

src/containers/internal/views/recent-calls/index.tsx
khanghugo: So I have a generic enough filter, I think that can be used for these too. But I am not sure how it would go without your magic touch.

src/containers/internal/views/recent-calls/index.tsx
khanghugo: I see that in the RecentList as code but I don't see it visually so I am not really sure how to test it. But I will add it accordingly for parity from the code.

kitajchuk: In the old UI you can toggle a line item and it shows all the call data as key/value pairs. The pcap can be downloaded from this state of it exists. The condition for if it exists is to make a fetch to the call by ID api and of the response status is 200 and the response json has a total field greater than zero you can then fetch the pcap blob and provide a download link/button for the user. This file is used in an application called wire shark where you can view the data from the pcap.

khanghugo: Alright, I got the expanded row in. Now I just need to add the pcap thing.

It looks really terrible right now because I just use the `details` tag to wrap the thing.

src/containers/internal/views/recent-calls/index.tsx
kitajchuk: Lol magic 🪄😅. When you’re ready drop a comment on the main thread and I’ll check this out and do a UI pass on it and see where we land. Once we wrap this one up tag team the alerts view will be cake 🍰 for you because it’s basically the same but simpler with only two data columns and no toggle etc.

src/containers/internal/views/recent-calls/index.tsx
kitajchuk: Cool all good. I’ll drop some UI magic 🪄 on it when you’re ready for it 😅🤘

khanghugo:
The Pcap thing is here. After some challenges of the refresh webapp requesting 50,000 requests, I just settled with doing exactly what the webapp is doing and it works well. Though, I am not sure why the webapp does not show the Pcap button so I can monitor how exactly it behaves. But it is now exactly what it is supposed to be.

kitajchuk:
> The Pcap thing is here. After some challenges of the refresh webapp requesting 50,000 requests, I just settled with doing exactly what the webapp is doing and it works well. Though, I am not sure why the webapp does not show the Pcap button so I can monitor how exactly it behaves. But it is now exactly what it is supposed to be.

50,000!!? Lol—sounds like you had some bad react state going on there.

khanghugo:
> 50,000!!? Lol—sounds like you had some bad react state going on there.

I was fetching pcap without fetching the recent call again and it happens. But then I just gave up and just copying the webapp and it works. Not sure why but that works. 

Right now I just hope what in this branch is exact to what is expected. It looks like it to me.

kitajchuk:
> > 50,000!!? Lol—sounds like you had some bad react state going on there.
> 
> I was fetching pcap without fetching the recent call again (fetch pcap in recent call fetch's `then()`)and it happens. But then I just gave up and just copying the webapp and it works. Not sure why but that works.
> 
> Right now I just hope what in this branch is exact to what is expected. It looks like it to me.

The api is very particular about this and the mock server replicates the backend. You have to fetch the recent call by call ID and that response just returns a total property and if it’s value is greater than zero that means the call has a pcap file, otherwise it doesn’t. So there’s no way to know if a call has a pcap if you are looking at the call data from the paged response.

khanghugo:
I am really trying to get the pagination to show up button to page 3 when at page 1 and conversely at the other end but I am doing too much of boolean stuffs and the prettifier (Prettier) keeps removing the parantheses that matter, lol. I mean that is me just doing this with just boolean so maybe there's that. I can definitely look up online and see how people do these things.

kitajchuk:
> I am really trying to get the pagination to show up button to page 3 when at page 1 and conversely at the other end but I am doing too much of boolean stuffs and the prettifier (Prettier) keeps removing the parantheses that matter, lol. I mean that is me just doing this with just boolean so maybe there's that. I can definitely look up online and see how people do these things.

I can help. Probably tomorrow / this week I can take a look. Just let me know when you’re ready to hand it off and I’ll do the UI work and check out the pagination stuff.

khanghugo:
From your feedback of the last reviews, I have added everything and reviewed everything. I don't think there is anything more to add on feature wise. It is still not a table because I am not sure how it would go but all the stuffs are there and I think you can wrap them nicely into a table just fine.

I am ready to hand it off.

kitajchuk:
> From your feedback of the last reviews, I have added everything and reviewed everything. I don't think there is anything more to add on feature wise. It is still not a table because I am not sure how it would go but all the stuffs are there and I think you can wrap them nicely into a table just fine.
> 
> I am ready to hand it off.

Cool cool 😎. I’ll peep it and see what I can do! I have some fun ideas 🙂

kitajchuk:
@khanghugo @davehorton I've done some good UI work on this one and refactored a bit so the extra pcap fetches only happen once per item if toggled 👍. The pagination is sticky positioned at the bottom so it's always present. The list feels much better to me now. Anyway, thoughts etc. Hopefully these styles are okay :)

Here's a screenshots of the updated UI:

Screen Shot 2022-09-11 at 2 25 02 PM
Screen Shot 2022-09-11 at 2 25 23 PM
Screen Shot 2022-09-11 at 2 25 49 PM

khanghugo:
That looks great. I really like that it is consistent with how over views look. But there is this one discussion about tabular view. Dave will have a say on this ultimately anyway. 

For me, because we are trying to get that thing consistent, each item seems to take more space. For 12 entries, I have to zoom out 50%, which is barely visible anything, to see all the entries. I think if we roll with this style then we need to reduce the spacing. I think we don't need the second line of info, that can reduce to one line, for a starter. 

What do you think?

kitajchuk:
> That looks great. I really like that it is consistent with how over views look. But there is this one discussion about tabular view. Dave will have a say on this ultimately anyway.
> 
> For me, because we are trying to get that thing consistent, each item seems to take more space. For 12 entries, I have to zoom out 50%, which is barely visible anything, to see all the entries. I think if we roll with this style then we need to reduce the spacing. I think we don't need to seperate the second line of info, that can make that one, for a starter.
> 
> What do you think?

To be fair the current webapp doesn't show all entries at once, even with the table you have to scroll. It does show up to 12 on an average laptop screen, though. I would be inclined to consider reducing the vertical padding overall. All text on one line is a responsive nightmare, but we can have breakpoints if we think we need to go that route and drop the line at some point when it needs to. The current webapp is not responsive at all and the UI becomes borked heavily around standard tablet screen sizes. Other things we could do are reduce the vertical padding between the page titles and the filter areas and drop the page title font-size as well. Reduce the padding on list items and do those things and that would help as well.

kitajchuk:
> > That looks great. I really like that it is consistent with how over views look. But there is this one discussion about tabular view. Dave will have a say on this ultimately anyway.
> > For me, because we are trying to get that thing consistent, each item seems to take more space. For 12 entries, I have to zoom out 50%, which is barely visible anything, to see all the entries. I think if we roll with this style then we need to reduce the spacing. I think we don't need to seperate the second line of info, that can make that one, for a starter.
> > What do you think?
> 
> To be fair the current webapp doesn't show all entries at once, even with the table you have to scroll. It does show up to 12 on an average laptop screen, though. I would be inclined to consider reducing the vertical padding overall. All text on one line is a responsive nightmare, but we can have breakpoints if we think we need to go that route and drop the line at some point when it needs to. The current webapp is not responsive at all and the UI becomes borked heavily around standard tablet screen sizes. Other things we could do are reduce the vertical padding between the page titles and the filter areas and drop the page title font-size as well. Reduce the padding on list items and do those things and that would help as well.

@khanghugo I pushed a commit that reduces vertical spacing between main sections and drops reduces the font-size of all page titles. In general I think this feels better but we're still left to solve overall list item sizing if we're talking rows per page visibility on an average laptop sized screen as a baseline 🤔

khanghugo:
Yea, moving the filters to one side seems nice. I am not sure what the baseline would be. Let see how Dave thinks as he was really mentioning this.

davehorton:
Thanks guys, this looks lot better to me. The only thing I noticed now was that in mobile phone view the summary line splits the phone numbers weirdly:

image

Any chance that we can fix that?  Perhaps even creating more space by dropping the icons?

davehorton (dismissed):

kitajchuk:
> Thanks guys, this looks lot better to me. The only thing I noticed now was that in mobile phone view the summary line splits the phone numbers weirdly:
> 
> image
> 
> Any chance that we can fix that? Perhaps even creating more space by dropping the icons?

Yup I can address that. There’s a couple other general mobile padding things I wanted to address as well so I’ll push one more commit here and ping you when you can prep it again 👍

Issue #76, opened 2022-08-10
Refresh: Alerts view


kitajchuk:
Implement functional parity for the `Alerts` view form with the new UI style.

Create a branch from `refresh` called `refresh-alerts` and submit a PR to the `refresh` branch. Please include this issue number (`#76`) in the commit/PR message.

You can find the container stub for this view [here](https://github.com/jambonz/jambonz-webapp/tree/refresh/src/containers/internal/views/alerts). It already has some code stubbed with mock data from a local dev server API implementation.

For any questions about design/UI/coding ping @kitajchuk.

Ultimately, when merging PRs, please always use "Squash and Merge" for the strategy.

kitajchuk:
@khanghugo A functional dev server for this view is now complete as well so both this and recent calls are wide open for you to jam on 👍

Commit: https://github.com/jambonz/jambonz-webapp/pull/64/commits/f348da6709346bfeb3249593302ec2372e3336a0

kitajchuk:
@khanghugo Recent calls has been merged so you can go ahead and utilize the cleaned up UI for this view 👍

PR #99, opened 2022-09-13
Alert view page for refresh 

khanghugo:
#76 

A really quick copy paste work.

No style for the table, as always.

I am sort of concerned of this `useEffect()`. It seems really adventurous.

https://github.com/jambonz/jambonz-webapp/blob/c29087ac9ecae0dde5682e2f2018325790bac8ae/src/containers/internal/views/alerts/index.tsx#L63-L72

Seems like we can call the `handleFilterChange()` in the former `useEfflect()` also?

### DIFF
--- a/src/containers/internal/views/alerts/index.tsx
+++ b/src/containers/internal/views/alerts/index.tsx
@@ -61,10 +61,6 @@ export const Alerts = () => {
   };
 
   useEffect(() => {
-    setPageNumber(1);
-  }, [perPageFilter]);
-
-  useEffect(() => {
     if (accountSid) {
       handleFilterChange();
     }
@@ -98,9 +94,7 @@ export const Alerts = () => {
               </div>
             ))
           ) : (
-            <div className="item">
-              <M>No data</M>
-            </div>
+            <M>No data</M>
           )}
         </div>
       </Section>
@@ -120,6 +114,7 @@ export const Alerts = () => {
             id="page_filter"
             filter={[perPageFilter, setPerPageFilter]}
             options={perPageSelection}
+            handleSelect={() => setPageNumber(1)}
           />
         </ButtonGroup>
       </footer>
--- a/src/containers/internal/views/alerts/index.tsx
+++ b/src/containers/internal/views/alerts/index.tsx
@@ -12,6 +12,7 @@ import {
   Section,
   SelectFilter,
   Spinner,
+  Icons,
 } from "src/components";
 
 import type { Account, Alert, PageQuery } from "src/api/types";
@@ -68,7 +69,7 @@ export const Alerts = () => {
           id="date_filter"
           label="Date"
           filter={[dateFilter, setDateFilter]}
-          options={DATE_SELECTION}
+          options={DATE_SELECTION.slice(0, 2)}
         />
       </section>
       <Section {...(hasLength(alerts) ? { slim: true } : {})}>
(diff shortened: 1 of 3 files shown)


kitajchuk:
@khanghugo look at the implementation I did for the recent calls view and you can use that approach here. TLDR; you only need one useEffect to fetch data, omit the perPageFilter as a dep and use the new handleSelect prop to reset the pageNumber on the per page SelectFilter component.

src/containers/internal/views/alerts/index.tsx
```diff
@@ -1,45 +1,123 @@
...
 
-    return function cleanup() {
-      ignore = true;
-    };
-  }, []);
+  useEffect(() => {
+    if (accountSid) {
+      handleFilterChange();
+    }
+  }, [accountSid, pageNumber, dateFilter, perPageFilter]);
```
kitajchuk, line 67: You can omit the perPageFilter since setting the page number in the handleSelect will trigger this effect 🤔

src/containers/internal/views/alerts/index.tsx
```diff
@@ -1,45 +1,123 @@
...
+          filter={[dateFilter, setDateFilter]}
+          options={dateSelection}
+        />
+      </section>
+      <Section {...(hasLength(alerts) ? { slim: true } : {})}>
+        <div className="list">
+          {!hasValue(alerts) && <Spinner />}
+          {hasLength(alerts) ? (
+            alerts.map((alert) => (
+              <div key={`${alert.alert_type}-${alert.time}`}>
```
kitajchuk, line 91: You can copy the nanoid usage I added to the dev server for recent calls so you don’t have to do this weird key with the mock data 👍

src/containers/internal/views/alerts/index.tsx
```diff
@@ -1,45 +1,123 @@
...
+  const [dateFilter, setDateFilter] = useState("today");
+
+  const [pageNumber, setPageNumber] = useState(1);
+  const [perPageFilter, setPerPageFilter] = useState("25");
+  const [maxPageNumber, setMaxPageNumber] = useState(1);
+
+  const [alerts, setAlerts] = useState<Alert[]>();
+  const [alertsTotal, setAlertsTotal] = useState(0);
+
+  const dateSelection = [
```
kitajchuk, line 30: Maybe we can use an api constant for this since it is the same here and for recent calls 🙂

src/containers/internal/views/alerts/index.tsx
```diff
@@ -1,45 +1,123 @@
...
+    { name: "last 14d", value: "14" },
+    { name: "last 30d", value: "30" },
+  ];
 
-    getAlerts("foo-sid", {
-      page: 1,
-      count: 25,
-      start: dayjs().startOf("date").toISOString(),
-    })
+  const perPageSelection = [
```
kitajchuk, line 37: Same thing here we could also make this an api constant 👍

src/containers/internal/views/alerts/index.tsx
kitajchuk: Nvm I’m wrong. There’s no entity sid for an alert 🤔. In that case what you have here is fine lol. If you’re curious the data for these last two views comes from this jambonz time series package: https://github.com/jambonz/time-series/—pretty neat 🙂

khanghugo: Hmmm, I don't really know what is the use for `key` attribute beside not making the intepreter whine or just to give some scraper dudes easier time.

src/containers/internal/views/alerts/index.tsx
khanghugo: I will be making changes related to that api constant and this view only. I can make a reminder in the post to make similar changes to recent calls.

src/containers/internal/views/alerts/index.tsx
kitajchuk: Heh—it’s purely in the react context and is not rendered. Read up here: https://reactjs.org/docs/lists-and-keys.html. You can create performance issues and rendering glitches if you use unstable/non-unique keys for lists. So the linting errors you get are for good reason 🤔

davehorton (approved):

Issue #100, opened 2022-09-13
Refresh: Pagination showing at least a number of button when at sides' ends


khanghugo:
Right now, when selected page is 1, it will show only 2, no 3 or 4. Hence, the number of shown buttons will move around and it can get annoying for that first few. I did really try to implement this with just boolean but the prettifier does not like what I do so I guess we can do something a bit more imperative for that.

kitajchuk:
@khanghugo Not a bad call. Currently it functions quite similar to the old webapp but it could be nice to make it jump around less on the user 🤔

kitajchuk:
@khanghugo Another enhancement issue could be to support the query parameters in the URL address such that you could technically share a page of results and it would do an initial fetch with all the correct filters and pagination criteria 🤔

PR #101, opened 2022-09-14
Refresh enhancement pagination logic

khanghugo:
#100 

This is what I really wanted to get done when I was working on the recent calls.

It is not quite showing at least how many button as I realize it would introduce some really complex logic. This would feel better than what the pagination is now.

### DIFF
--- a/src/components/pagination/index.tsx
+++ b/src/components/pagination/index.tsx
@@ -15,6 +15,7 @@ type PaginationProps = {
 
 const nextTo = 1;
 const jumpNum = 3;
+const showMin = 4; // hard coded and tested for now, if we want something more sophisticated, sure
 
 export const Pagination = ({
   pageNumber,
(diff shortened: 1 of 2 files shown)


khanghugo:
Hmmm, wrong HEAD.

khanghugo:
Maybe I can divert that to refresh-alert instead.

khanghugo:
hmmm, even wronger commit.

khanghugo:
@kitajchuk

src/components/pagination/index.tsx
```diff
@@ -15,6 +15,7 @@ type PaginationProps = {
 
 const nextTo = 1;
 const jumpNum = 3;
+const showMin = 4; // hard coded and tested for now, if we want something more sophisticated, sure
```
khanghugo, line 18: Just a quick note for this, this doesn't mean it doesn't work with different value, it just feels nicer for the current nextTo value with this value.

kitajchuk: I toy with the idea in my mind but since this is the only place they are used it's definitely fine to have them declared here like this. Any number of options for refactoring later (if ever need be) whether it's constants from a higher level file or accepting these values as props with defaults 🤔

kitajchuk: @khanghugo I just pushed a commit that makes these optional props. I think this is fair so we have sane defaults for them but makes the component more flexible downstream if we ever wanted to set different values for some reason.

kitajchuk:
@khanghugo Check my latest commit/comment. I think this one is ready from my POV. It feels nice. Unless you're working on anything more with this I think we're good.

khanghugo:
Yea, that one is all set from me.

PR #106, opened 2022-09-19
Fuzzy search filter for collection lists

kitajchuk:


kitajchuk:
@davehorton Give this a test drive. It feels pretty good. I abstracted the main filter logic into a custom hook that uses memoization etc. The fuzziness allows searching against any "string"-able field on a data object so one thing that is also neat is you can just paste a SID into the field and get just that one item instantly. I implemented this feature on Accounts, Applications, Carriers, Phone Numbers and MS Teams Tenants.

Here's some screenshots:

Screen Shot 2022-09-19 at 11 57 24 AM


Screen Shot 2022-09-19 at 12 00 55 PM

Screen Shot 2022-09-19 at 12 01 32 PM

khanghugo:
Looking sharp.

kitajchuk:
> Looking sharp.

lol thanks 😅

davehorton (commented):
just wondering...which fields does the fuzzy search work against?  For account I see it works against name, does it work against any other properties of account?  Same question for carrier and phone number etc..

davehorton:
OK I see it works for account sid as well, I just needed to enter at least 3 or so digits.  That's good

kitajchuk:
> OK I see it works for account sid as well, I just needed to enter at least 3 or so digits. That's good

Ya it works for any string-able field generically 🤔

davehorton (dismissed):

Issue #104, opened 2022-09-16
Bug: Phone numbers can be assigned an application that does not belong to the same account that the number/carrier belongs to


kitajchuk:
You should only be able to assign an application to a phone number where the application belongs to the same account that the phone number/carrier belongs to (with the exception that the carrier can be associated to the account’s parent service provider).

Bug: currently you can assign a phone number to an application that does not belong to the same account that the phone number/carrier belongs to via the mass edit function from the phone numbers list view.

Proposed solution: provide an account filter to the phone numbers list view and only further provide the mass edit UI if an account has been selected to filter the list. This means we would need to maintain a filtered set of phone numbers as our primary render list and the application selector would need to filter applications by the selected account.

xquanluu:
Start working on this

kitajchuk:
@xquanluu I was meaning to document some implementation details for this I just hadn't gotten around to it. We essentially have examples of exactly how to make the changes we need to see here already on the carriers index so I'll document the necessary changes here so that we keep things DRY and follow like patterns. Hopefully you're not too far in to anything else or maybe you already stumbled on these pre-existing patterns.

First you'll want to use the `AccountFilter` component like we do on other views. Here's the example from the carriers index:
https://github.com/jambonz/jambonz-webapp/blob/refresh/src/containers/internal/views/carriers/index.tsx#L110-L115

This phone numbers view already has a fuzzy filter so it will need 2 layers of filter now: first by `accountSid` and then by fuzzy search filter. We have already done this on the carrier view as well so the implementation here should be the same:
https://github.com/jambonz/jambonz-webapp/blob/refresh/src/containers/internal/views/carriers/index.tsx#L34-L47

With that in place all you'll need to do is conditionally render this first list item so long as the `accountSid` is set:
https://github.com/jambonz/jambonz-webapp/blob/refresh/src/containers/internal/views/phone-numbers/index.tsx#L120-L169

I think that's pretty much it. This should be an easy one since we just need to do the above. If I'm missing anything let me know, but I'm pretty sure this is it.

khanghugo:
Hmmm, this was what I was talking about in the PR thread too, I believe. The old webapp does behave this way so I did have it in parity. Glad to see it gets clarified.

PR #108, opened 2022-09-20
fix: #104 Phone Number can be assgined to the same account's application

xquanluu:


xquanluu:
#104

src/containers/internal/views/phone-numbers/index.tsx
```diff
@@ -91,6 +96,26 @@ export const PhoneNumbers = () => {
     }
   };
 
+  useMemo(() => {
+    setPhoneNumbers(
```
kitajchuk, line 100: These patterns are both incorrect. You should not set state in a memo callback. Memo is designed for very specific use cases. The memo callback should return the value that is assigned to the memoized variable. You actually don’t need either of the new state vars you created here and can go back to the original assignments with the use api hooks. I linked to an example of exactly how this should be implemented in the issue. TBH it doesn’t even look like this code even works because the view is still using the filteredPhoneNumbers to map and render… I would suggest reading that implementation details I posted on this issue thoroughly.

kitajchuk (changes requested):
Please read the implementation details I posted on the issue for this bug. As it stands this code has generally incorrect usage of memoization and from reading it doesn’t even appear it would produce the desired result.

My notes on implementation: https://github.com/jambonz/jambonz-webapp/issues/104#issuecomment-1251733892

xquanluu:
Hi @kitajchuk  @davehorton I have merged PR from @kitajchuk  for fixing this issue to this PR, please help me to review it again here, I tested local with new change and confirmed thing is working as expected.

kitajchuk:
> Hi @kitajchuk @davehorton I have merged PR from @kitajchuk for fixing this issue to this PR, please help me to review it again here, I tested local with new change and confirmed thing is working as expected.

Heh. I left a comment on that PR for you. Hopefully it clears up my key point here which is understanding react and it's APIs and how/when to use things. As for this PR, we just need to nuke the changes in the `.gitignore` since they don't belong here. I can go ahead and do that...

kitajchuk:
@xquanluu I've nuked the `.gitignore` changes. Funny enough to your point on the other thread about less code—this is nearly exactly the same amount of code changes (additions/subtractions) as your original PR with the memo/state issues.

Screen Shot 2022-09-20 at 4 14 30 PM

The fundamental difference is in how we're using memo for the filtered set here. Rather than setting state twice in two loose memo calls (not good—anti-pattern) we're assigning a single variable to a memoized result:

```tsx
const phoneNumbersFiltered = useMemo(() => {
    return phoneNumbers
      ? phoneNumbers.filter(
          (phoneNumber) => !accountSid || phoneNumber.account_sid === accountSid
        )
      : [];
  }, [accountSid, phoneNumbers]);
```

The other critical difference here is it means we're not using unnecessary state. The two extra state vars you added are not needed because `phoneNumbers` and `applications` are fetched with our api hooks. The filtered phone numbers are managed via memo and the addition of our fuzzy filter results hook and the applications can just be filtered inline based on presence of the `accountSid` or not. So we get all the changes we need with a single state var: `accountSid`.

PR #110, opened 2022-09-20
Working solution to issue #104

kitajchuk:
@xquanluu It's probably tough to totally explain clearly what I'm talking about in my PR review for #104. Here's a commit pointed back at your branch showing the intended implementation for this bug to be fully functional with regard to the proposed solution of requiring account selection before you can assign applications.

Here's also some screenshots of this branch in action:

Screen Shot 2022-09-20 at 1 16 47 PM

Screen Shot 2022-09-20 at 1 17 04 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,9 @@ dist
 dist-ssr
 *.local
 yarn.lock
+
 # Editor directories and files
 .idea
-.vscode
 .DS_Store
 *.suo
 *.ntvs*
--- a/src/containers/internal/views/phone-numbers/index.tsx
+++ b/src/containers/internal/views/phone-numbers/index.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo, useState } from "react";
-import { Button, ButtonGroup, H1, Icon } from "jambonz-ui";
+import { Button, ButtonGroup, H1, Icon, MS } from "jambonz-ui";
 import { Link } from "react-router-dom";
 
 import {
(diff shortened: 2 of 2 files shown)


xquanluu:
Thanks @kitajchuk I was trying to reduce number of changes in code we need for the bug, my PR is also showing the same result as yours. But maybe there is some issue that I don't know yet. I will decline my PR, please go ahead with your PR

kitajchuk:
> Thanks @kitajchuk I was trying to reduce number of changes in code we need for the bug, my PR is also showing the same result as yours. But maybe there is some issue that I don't know yet. I will decline my PR, please go ahead with your PR

The main issue here is that we ought not set state in a use memo callback. Here's docs on use memo:

https://reactjs.org/docs/hooks-reference.html#usememo

Use memo should return a value, not set state or perform any side effects. So in this case, you are setting state in the use memo callback which is an anti-pattern for react. So, rather, we ought to compute some logic and return the value. So calls to use memo should be assigned to a variable like I have here.

Sure less code is a good goal (though your original PR isn't actually much less code than this TBH). But it needs to be correct code. The diff here looks bigger but it's only because a chunk of HTML was wrapped in a render condition.

The results are not _exactly_ the same. I think the phone numbers list should not be filtered by an account by default or as the only way to view them. For this we use the `defaultOption` on the `AccountFilter` so that you can see everything. 👈 This stuff is nuance to overall UX.

The main point to drive home here is the memo/state issues we see. React is interesting because it's somewhat approachable in general. However, it's easy to muck up in these ways so my point is really about understanding react and it's APIs and how/when to use them. Hopefully this makes sense.

Issue #109, opened 2022-09-20
Feature: ability to specify an optional call limit at the account and service provider level


davehorton:
There is a new feature we have added to the API that allows the system to limit the number of concurrent calls allowed for a given account and/or service provider.  To support this, there are two new tables:

```
CREATE TABLE service_provider_limits
(
service_provider_limits_sid CHAR(36) NOT NULL UNIQUE ,
service_provider_sid CHAR(36) NOT NULL,
category ENUM('api_rate','voice_call_session', 'device') NOT NULL,
quantity INTEGER NOT NULL,
PRIMARY KEY (service_provider_limits_sid)
);
CREATE TABLE account_limits
(
account_limits_sid CHAR(36) NOT NULL UNIQUE ,
account_sid CHAR(36) NOT NULL,
category ENUM('api_rate','voice_call_session', 'device') NOT NULL,
quantity INTEGER NOT NULL,
PRIMARY KEY (account_limits_sid)
);
```
> Note: although we allow for limiting number of api calls/min and number of devices registered, this is not currently implemented in the API - only the limit of number of calls (voice_call_session) is implemented for now.

There is new associated APIs:
```
POST /ServiceProviders/:sid/Limits
{category: 'voice_call_session', quantity: 100}

GET /ServiceProviders/:sid/Limits

and similarly

POST /Accounts/:sid/Limits
{category: 'voice_call_session', quantity: 100}

GET /Accounts/:sid/Limits
```
swagger doc [here](https://github.com/jambonz/jambonz-api-server/blob/052a19cfdc36acb50d890d6b1d9e69b786f7212a/lib/swagger/swagger.yaml#L1858) and [here](https://github.com/jambonz/jambonz-api-server/blob/052a19cfdc36acb50d890d6b1d9e69b786f7212a/lib/swagger/swagger.yaml#L1916).

> Note: the POST will update an existing limit if one exists, otherwise create one.

What is needed now is the ability to view and set the limits in the jambonz webapp.  We need this in both the "legacy" webapp (main branch) as well as the "refresh" branch, but the priority is the main branch.

On the Settings page (which is about the service provider) we should add:
- an input field labeled "Max Concurrent Calls"

Similarly on the Accounts page
- an input field labeled "Max Concurrent Calls"

The value of the field shall be populated using the GET apis above.  If the user enters a value, then we should populate to the database using the POST API above.

Note: a value of 0 is taken to mean no limit will be applied.

davehorton:
[...]

PR #112, opened 2022-09-21
feat: add Service provider limits and account limits to settings

xquanluu:


xquanluu:
#109 @kitajchuk  @davehorton  Please help me to review the PR for limits feature on main branch

davehorton:
I see some errors in the console, see below

account page
image

and settings page

image

davehorton (commented):
review errors on chrome console and fix

xquanluu:
@davehorton just fixed your comments.

davehorton:
still seeing this error or warning, which I think is new
image

davehorton:
also can you move this label and field down on the account page, to just above the line with "Subspace"?

and can we make the label say "Max Concurrent Calls (0=unlimited)"

xquanluu:
Thanks @davehorton for your deep check, I fixed it for that issue.

src/components/forms/AccountForm.js
```diff
@@ -633,6 +661,36 @@ const AccountForm = props => {
     },
   ];
 
+  const limitElements = [];
```
kitajchuk, line 664: This is a very strange pattern for react. Rather than do this you should map over the limits directly in the JSX tree and render each limit there.

src/components/forms/AccountForm.js
```diff
@@ -686,6 +744,7 @@ const AccountForm = props => {
           autoFocus={props.type === 'setup'}
           ref={refSipRealm}
         />
+        {props.type === 'edit' && limitElements}
```
kitajchuk: See notes on mapping instead.

src/components/forms/SettingsForm.js
```diff
@@ -208,15 +223,10 @@ const SettingsForm = () => {
...
-        method: 'put',
-        baseURL: APP_API_BASE_URL,
-        url: `/ServiceProviders/${serviceProviderSid}`,
-        headers: {
-          Authorization: `Bearer ${localStorage.getItem('token')}`,
-        },
-        data,
-      });
+      await callApi(`/ServiceProviders/${serviceProviderSid}`, 'put', data);
+      for (const limit of localLimits) {
```
kitajchuk, line 227: Again forEach is preferable here

src/components/forms/SettingsForm.js
```diff
@@ -253,6 +263,36 @@ const SettingsForm = () => {
     }
   };
```
kitajchuk, line 265: Also again use array map rather than this pattern.

src/components/forms/SettingsForm.js
```diff
@@ -253,6 +263,36 @@ const SettingsForm = () => {
...
+    limitElements.push(
+      <Input
+        name={`input-${category}`}
+        id={`input-${category}`}
+        key={`input-${category}`}
+        type="number"
+        min="0"
+        value={quantity ? quantity : 0}
+        onChange={e => {
+          let isLimitExisted = false;
```
kitajchuk, line 279: I see for this PR you are kind of trying to riff from my implementation on the refresh branch. In that case this onChange logic is overdone a bit. See what I did here:

https://github.com/jambonz/jambonz-webapp/blob/3b28d539d9fedb661d962a0f7bead07c9e92de19/src/components/local-limits.tsx#L52

kitajchuk: Looks like the difference is because I set an initial default value for the state var so maybe you’re okay with what you have here. I asked @davehorton if having an initial default is okay or not so I may have to change that on the refresh branch 🤔. Seems safe enough to have defaults since 0 equals no limit but we’ll see 🤷‍♂️

kitajchuk (changes requested):
I think there’s some desirable changes to be made here. See comments.

src/components/forms/SettingsForm.js
kitajchuk: Looks like no to default so ignore me on this one 🙃

src/components/forms/AccountForm.js
xquanluu: This is because the map will need to return a single <div> tag incluse 2 elements Label and Input.
This causing the layout is moved to the left as the legacy layout is designed for 1 label on the left, 1 Input on the right

may be you have some better suggestion

kitajchuk: Yes: use a fragment to wrap the two elements so you get the desired result. Either `<>...</>` or `<React.Fragment>...</React.Fragment>`.

src/components/forms/SettingsForm.js
xquanluu: foreach is not able to call await to make sure the call API is finished, Can you give a suggestion?

kitajchuk: I don't know that I see why you need to use await here 🤔. You're not assigning the response data from the api call so it seems unnecessary TBH. This old application is littered with a lot of gross async/await 🙃.

src/components/forms/AccountForm.js
```diff
@@ -892,6 +920,38 @@ const AccountForm = props => {
           </Button>
         )}
 
+        {props.type === 'edit' && LIMITS.map(({ label, category }) => {
```
kitajchuk, line 923: I think we should let them add these for add and edit 👍

package-lock.json
```diff
@@ -13,7 +13,7 @@
         "@testing-library/user-event": "^7.2.1",
         "antd": "^4.21.0",
         "axios": "^0.21.1",
-        "moment": "^2.29.1",
+        "moment": "^2.29.4",
```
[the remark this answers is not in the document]
kitajchuk: I suppose you can leave this in here. Looks like the folks that updated moment neglected to check in the package lock file which is why you ended up with these changes. Not your fault lol 🤷

src/components/forms/AccountForm.js
```diff
@@ -892,6 +920,38 @@ const AccountForm = props => {
...
+          return <React.Fragment key={`fragment-${category}`}>
+            <Label htmlFor={`label-${category}`} key={`label-${category}`}>{label}</Label>
+            <Input
+            name={`input-${category}`}
+            id={`input-${category}`}
+            key={`input-${category}`}
+            type="number"
+            placeholder={label}
+            min="0"
+            value={quantity ? quantity : ""}
```
kitajchuk, line 934: Seems to be an inconsistency here because if quantity is 0 then this renders an empty value. However, an empty value ends up being 0 because `Number("") === 0`. So, the question is do we render the 0 or leave this blank. Either way the 0 implies no limit on the backend, however to the user it looks like no limit is set when in fact a limit of 0 is set and returned from the backend.

kitajchuk: Also this is weird because 0 if falsey so I cannot manually enter a 0 into the field...

davehorton (approved):

davehorton:
looks good to me in testing.  @kitajchuk are your comments above addressed?

kitajchuk:
> looks good to me in testing. @kitajchuk are your comments above addressed?

I can check in a bit. I was hoping if we’re ready to merge the refresh branch that we could work that out first and then merge this PR into the legacy branch. Saves me one last rebase with main on the refresh branch 🤔. But I’m fine with whatever we need to do.

.gitignore
```diff
@@ -15,6 +15,7 @@
 .env.development.local
 .env.test.local
 .env.production.local
+.env
```
kitajchuk, line 18: We intentionally don’t ignore this base .env file. This needs to be removed from this PR. I can’t stress enough how important it is to not clutter feature PRs with random changes like this—especially flat out incorrect ones. Much like the vscode configs on the refresh branch we have this file checked into git for a reason. It’s even referenced in the readme.

Everything in this PR needs to be for the feature in question, anything else doesn’t belong here.

.env
```diff
@@ -1,2 +0,0 @@
-REACT_APP_API_BASE_URL=http://127.0.0.1:3002/v1
```
kitajchuk, line 1: Take this out and see my other comment. You should not delete this file. Seriously throwing changes like this into a feature PR is poor practice. Please keep the PRs properly in scope.

src/components/forms/AccountForm.js
```diff
@@ -319,6 +320,16 @@ const AccountForm = props => {
             },
           });
           promiseList.push(applicationsPromise);
+
+          const limitsPromise = axios({
+            method: 'get',
+            baseURL: APP_API_BASE_URL,
+            url: `/Accounts/${[props.account_sid]}/Limits`,
```
kitajchuk, line 327: You can remove the square brackets as they are completely unnecessary.

kitajchuk (changes requested):
So I have some things I want removed from this PR. See my comments.

kitajchuk:
@xquanluu @davehorton I've created the legacy branch and cleaned up the stale branches. I updated this PR to point at the legacy branch so the main branch is pencils down until I merge the refresh branch.

kitajchuk:
Oh crud, @xquanluu I forgot to pull latest on this branch before I updated it with that rebase. You'll need to get your latest changes back up here. Here's how:

```
# from your #109 branch
git fetch --all
git pull origin legacy --rebase
git push origin #109 -f
```

This should do it 🤔

src/components/forms/AccountForm.js
```diff
@@ -567,6 +581,23 @@ const AccountForm = props => {
         },
         data: axiosData,
       });
+      // Update Limits
+      const acc_sid = accountSid ? accountSid : accountResp.data.sid;
+      await Promise.all(
+        localLimits.map(l => {
+          const method = l.quantity ? 'post' : 'delete';
```
kitajchuk, line 588: I think you should use the empty string check here otherwise if they enter a zero you will delete the limit and it will result in rendering an empty string when the data updates which would be unexpected. Entering a zero should be valid for posting, not deleting. I see in the SettingsForm you used the empty string check so you can do it like that here as well.

src/components/forms/AccountForm.js
```diff
@@ -567,6 +581,23 @@ const AccountForm = props => {
...
+          const limitUrl = l.quantity ? `/Accounts/${acc_sid}/Limits` : `/Accounts/${props.account_sid}/Limits?category=${l.category}`;
+          const obj = Object.assign({
```
kitajchuk, line 590: There is no need to use `Object.assign` here. In fact, you could just pass the object directly into your `axios` call and return that to clean this up a bit. Like this:

```js
return axios({
  method: method,
  baseURL: APP_API_BASE_URL,
  url: limitUrl,
  headers: {
    Authorization: `Bearer ${jwt}`,
  },
  data: method === 'post' ? l : {},
});
```

src/components/forms/AccountForm.js
```diff
@@ -892,6 +923,31 @@ const AccountForm = props => {
...
+          return <React.Fragment key={category}>
+            <Label htmlFor={category}>{label}</Label>
+            <Input
+            name={category}
+            id={category}
+            type="number"
+            placeholder="Enter Quantity (0=unlimited)"
+            min="0"
+            value={quantity >= 0 ? quantity : ""}
+            onChange={e => {
```
kitajchuk, line 937: Nice I see you took a look at how I did this in the new app with the `LocalLimits` component 👍

src/components/forms/SettingsForm.js
```diff
@@ -60,6 +60,19 @@ const SettingsForm = () => {
   const [serviceProviderSid, setServiceProviderSid] = useState('');
   const [serviceProviders, setServiceProviders] = useState([]);
   const [confirmDelete, setConfirmDelete] = useState(false);
+  const [localLimits, setLocalLimits] = useState([]);
+
+  const callApi = async (path, method, data) => {
+    const obj = Object.assign({
```
kitajchuk, line 66: Same thing as before—you don't need `Object.assign` here since you can just create the `obj` as a direct reference.

kitajchuk (changes requested):
Overall looking pretty good. I just have a couple comments on things that can be changed including one note about how you're checking `quantity` to determine post or delete method.

src/components/forms/AccountForm.js
```diff
@@ -585,17 +585,17 @@ const AccountForm = props => {
...
             method: method,
             baseURL: APP_API_BASE_URL,
             url: limitUrl,
             headers: {
               Authorization: `Bearer ${jwt}`,
-            }
-          }, method === 'post' ? {data: l} : {});
-          return axios(obj);
+            },
+            data: method === 'post' ? l : undefined
```
davehorton, line 597: a more concise way to optionally include a parameter using ES6 syntax that I favor is
```
...(method === 'post && {data: l})
```

kitajchuk: Ha that’s a good one. Yes this code is all compiled with babel so this is just fine 🙂

davehorton: @kitajchuk are all your comments resolved?  Can I merge this now?

src/components/forms/SettingsForm.js
```diff
@@ -60,6 +60,19 @@ const SettingsForm = () => {
...
+
+  const callApi = async (path, method, data) => {
+    return await axios({
+      method: method,
+      baseURL: APP_API_BASE_URL,
+      url: path,
+      headers: {
+        Authorization: `Bearer ${localStorage.getItem('token')}`,
+      },
+      ...(method === 'post' && {data: data})
```
kitajchuk, line 73: This is not going to send the data for “put” requests so you’ll need to update this logic.

kitajchuk (changes requested):
Looks like one more thing to fix real quick.

src/components/forms/SettingsForm.js
kitajchuk: @davehorton Not quite I noticed this callApi method is designed to not send data for “put” requests now… I think this might be the last thing to fix here. I’m on my phone tho so I admit I haven’t actually run these latest changes yet 🤔

kitajchuk (approved):

kitajchuk:
@xquanluu @davehorton this seems like it probably works now but I admit I’m on my phone and can’t test the runtime right now.

Issue #113, opened 2022-09-24
Bug: MS Teams uses incorrect Accounts and Applications


kitajchuk:
The MS Teams view needs to fetch Accounts and Applications by the Current Service Provider and the Application select menu needs to be filtered by the current selected Account.

This update to initial fetches needs to happen in the index list view and the form add/edit view.

PR #114, opened 2022-09-24
Fix MS Teams view bug for #113

kitajchuk:
See issue here: https://github.com/jambonz/jambonz-webapp/issues/113


src/containers/internal/views/phone-numbers/index.tsx
```diff
@@ -50,7 +50,7 @@ export const PhoneNumbers = () => {
   const phoneNumbersFiltered = useMemo(() => {
     return phoneNumbers
       ? phoneNumbers.filter(
-          (phoneNumber) => !accountSid || phoneNumber.account_sid === accountSid
+          (phn) => !accountSid || phn.account_sid === accountSid
```
kitajchuk, line 53: This is just a nit pick change since `phoneNumber` is the name of a state var declared above in this component.

PR #120, opened 2022-09-27
Legacy: Microsoft custom TTS and STT endpoints

xquanluu:


davehorton:
when doing a PUT to update, please do not include the api_key (for Azure) or the json key (for google), or the aws key or secret key, because these have been obfuscated and you are sending back the obfuscated key, which we then gets stored, overwriting the actual key

```
{"level":30, "time": "2022-09-27T15:23:09.425Z","pid":3151,"hostname":"MacBook-Pro-3.local","api_key":"79b268XXXXXXXXXXXXXXXXXXXXXXXXXX","region":"eastus","use_custom_tts":0,"custom_tts_endpoint":null,"use_custom_stt":0,"custom_stt_endpoint":null,"msg":"testing microsoft tts","v":1}
```

xquanluu:
> 

Fixed

davehorton:
hmm, I just tested latest and if I am editing an existing credential, and uncheck the checkboxes, the custom endpoints still show -- and they are still sent in the PUT payload when I click Save

```
{
	"vendor": "microsoft",
	"service_provider_sid": "2708b1b3-2736-40ea-b502-c53d8396247f",
	"account_sid": null,
	"use_for_tts": 1,
	"use_for_stt": 1,
	"region": "eastus",
	"use_custom_tts": 0,
	"use_custom_stt": 0,
	"custom_tts_endpoint": "http://customvoice.com",
	"custom_stt_endpoint": "http://custom.stt"
}
```

kitajchuk:
> hmm, I just tested latest and if I am editing an existing credential, and uncheck the checkboxes, the custom endpoints still show -- and they are still sent in the PUT payload when I click Save
> 
> ```
> {
> 	"vendor": "microsoft",
> 	"service_provider_sid": "2708b1b3-2736-40ea-b502-c53d8396247f",
> 	"account_sid": null,
> 	"use_for_tts": 1,
> 	"use_for_stt": 1,
> 	"region": "eastus",
> 	"use_custom_tts": 0,
> 	"use_custom_stt": 0,
> 	"custom_tts_endpoint": "http://customvoice.com",
> 	"custom_stt_endpoint": "http://custom.stt"
> }
> ```
> 
> This might be just my misunderstanding of part of our discussion earlier on what happens when you check/uncheck/check the box. I was thinking of two different scenarios:
> 
> 1. you create and save a credential with a custom endpoint.  You then edit it.  The checkbox is checked and there is a custom endpoint string to start with.  Now you uncheck the checkbox.  Before you click Save, I was thinking that the custom endpoint field would now be empty.  But, if you immediately then click the checkbox to enable, the field would now again show the previous endpoint string -- i.e., it would have not been "lost".
> 2. As above, you create and then edit the credential. You uncheck the checkbox and then you do click Save.  I thought the payload would include use_custom_tts: 0 and custom_tts_endpoint: null
> 
> @kitajchuk I am not sure what you were thinking the behavior would be for those two scenarios, whether something different or as I have outlined. I'm actually fine at this point however you want it to work

@davehorton Hmm, yes. Originally in my prior comment with "Option 2" I was thinking this:

> If, as a user, I really don't want those fields anymore I can delete the values and save—but they are not deleted on my behalf if I save with the box un-checked.

However, I see your point here. So to me it sounds like you feel things should behave the way the Settings `checkzone` does right now, which is unique at the moment since it was the first one developed and we didn't necessarily have a solid POV on how this scenario should behave across the boards. The behavior on this one is like this:


https://user-images.githubusercontent.com/438711/193055994-a12202f9-79e7-40b9-b789-daaf6eaeeafe.mp4

Unfortunately the logic for this behavior requires holding on to some temp state in order to repopulate the field(s)—but this I think is only necessary when using optional fields that _are not hidden when the box is un-checked_. All of our other optional `checkzone` fields are hidden when un-checked. In my POV these custom TTS/STT fields could behave exactly the way this FQDN field does on the settings form. So to me it sounds like what we want is Option 2 (hold on to field values in case I check the box again)—but if I leave the box un-checked and save the fields should be set to `null` when sent to the backend. After I get the toast if I were to then toggle the zone open again the values would be empty since they were deleted on the backend. And obviously things would continue to function as they do—if the box is checked the fields are always kept and saved as expected.

@davehorton Does this sound like the right compromise here?

davehorton:
Yes, that does sound like the right option to me.  I think I may have over-complicated it, but if we could get the behavior that we see in your most recent video above, that is what I was thinking would be great.

xquanluu:
Update both PRs to have same behavior that use tmp state to store custom tts/stt endpoint when enable/disable it.

davehorton:
Looks great!  Thanks @xquanluu

PR #121, opened 2022-09-27
Main: Microsoft custom TTS and STT endpoints

xquanluu:


src/containers/internal/views/speech-services/form.tsx
```diff
@@ -91,8 +95,22 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
           currentServiceProvider.service_provider_sid || null,
         use_for_tts: ttsCheck ? 1 : 0,
         use_for_stt: sttCheck ? 1 : 0,
-        aws_region: vendor === VENDOR_AWS ? region : null,
-        region: vendor === VENDOR_MICROSOFT ? region : null,
+        ...(vendor === VENDOR_AWS && {
```
xquanluu, line 98: Without keys and passwords for vendor, cannot PUT update

davehorton: hmm, what do you mean?  The PUT fails with some kind of error?

The only thing we want to use the PUT for is to update the "use for tts" and "use for stt" settings.  Is that not working?

kitajchuk: The backend is presumably fixed for this so pull latest on the api server and it should be okay @davehorton ?

davehorton: yes could you pull latest main from jambonz-api-servr and test with that

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -180,6 +193,11 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
       if (credential.data.aws_region) {
         setRegion(credential.data.aws_region);
       }
+
+      setUseCustomTts(credential.data.use_custom_tts > 0 ? true : false);
+      setCustomTtsEndpoint(credential.data.custom_tts_endpoint || "");
```
kitajchuk, line 198: I think with null values returned from the backend we don’t need to set this state unless the value is not null 🤔. These endpoint state vars may be better wrapped in an if statement if that’s the case so we don’t needlessly call the setter. Have you tested with the updated backend yet?

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -355,6 +373,58 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
...
           </fieldset>
         )}
+        {vendor === VENDOR_MICROSOFT && (
+          <React.Fragment>
+            <fieldset>
+              <Checkzone
+                hidden
+                name="use_custom_tts"
+                label="Use for custom text-to-speech"
+                initialCheck={useCustomTts}
```
kitajchuk, line 383: In practice I can’t say I’m satisfied with using the payload state var as the value for the initialCheck prop. This causes redundant state updates since the Checkzone has an effect for this. I think now that we’ve had time with it the Checkzone would be better refactored to receive a state tuple for [checked, setChecked] more like we do with the generic selectors. But this change is out of scope here as it would require updating all usages of Checkzone to the new design 🤔

kitajchuk: I actually don’t even think that would solve it to be honest. The usages of this component somewhat complex in some cases and the point of the initialCheck prop is to be only to tell if the zone should initially be checked and open “on mount”. So ya, the model has been to use a unique state var just for that prop so that we don’t cascade unnecessary state applications.

kitajchuk: Actually I think it could work to switch to state tuple and not have initialCheck prop 🤔. Will have to explore it a bit on the side since it’s a generic change that’s out of scope here. But ya I think that could work lol

kitajchuk: Actually I think I wouldn't want to make this change to the Checkzone component because it's a bit of an anti-pattern to its design. The Checkzone is designed for things in which the dynamic checked state of the zone is not relevant to the use case. So we actually probably shouldn't use it here since we are using it incorrectly lol. Let's just switch to regular checkboxes like the other TTS/STT and have them conditionally render plain inputs if checked or not. That would make this much better :)

xquanluu: yes sure, and another problem is that the required prop is not getting effect when the field is important for submit.
I will change this and ask for review again.

src/containers/internal/views/speech-services/form.tsx
xquanluu: For my personal opinion, UI should not assume that backend is working all the time.
But applied the suggestion to the code.

xquanluu: I face with another usecase:
1/ From UI set use_custom_tts = 1, custom_tts_endpint="http://ttsendpoint.com", Submit the form and see toast Successfully created
2/ From UI set use_custom_tts=0, UI will update backend with custom_tts_endpint= null,  Submit the form and see toast Successfully updated
3/ UI still showing customTtsEndpoint = "http://ttsendpoint.com", even the API return null for custom_tts_endpint

So either we setCustomTtsEndpoint when checkbox for use_custom_tts is unchecked or setCustomTtsEndpoint to whatever we receive from GET method when refetch the speech credential.

I chose  setCustomTtsEndpoint to whatever we receive from GET method when refetch the speech credential.

kitajchuk: Ya I see what you're saying. Your original code was setting the field to empty string on change of the checkbox but I see you've changed that now. Personally, it feels weird to me to uncheck the box and still see my endpoint value in the field and then when I save it goes away. Sort of weird UI "magic" that feels strange. We do something like this on the Settings form with the MS Teams FQDN. In this case when you un-check the box it removes the text from the field and if you click the Save button it is removed on the backend. However, it keeps the field value in a temp state when you un-check the box so that if you check it again it is repopulated. I think a compromise would be nice here. If I un-check the box I want to visually see the field entry go away. Whether you want to just wipe it in the client session or use the temp state mechanic we have on the Settings form is up to you. It's a common mechanic of us to wipe the field if you un-check the optional thing. The Carrier form has this for tech prefix, diversion header etc. I think the temp state thing is unique to the Settings form and not actually the norm here. The Settings form was the first thing built after login in this new UI so it's the oldest model. I think it seemed fine at the time but many other forms have a lot of optional things and it seems unreasonable to manage temp state for any of them lol. I think the other mechanic that is in play in our more common scenarios is that the optional field is actually not rendered unless the box is checked—this is another approach you could consider here. Optional but disabled fields is actually not the norm so far in this new UI. This was also the case with the old web UI as well (hide/show on check) so users are used to it. It's nice because UI you're not using isn't taking up space visually as well. So maybe that would be a better approach here 🤔.

To your point about the UI not assuming the backend is working all the time—your compensation is that the field value doesn't go away then repopulate when the broken backend returns the old value, but this isn't really an issue to be concerned with. The backend _should_ work—that's the point. If it doesn't work the frontend should express a "bug" which it does either way in this case. The bug should be fixed and things work as expected.

davehorton: I agree with brandon on the last point -- if there is a bug in jambonz-api-server let's fix it, not work around it.  Let me know what the problem is and I will fix it (or feel free to take a crack at it yourself)

kitajchuk: Heh I think the issue is resolved now on the backend lol. For me the biggest question here is the UI/UX and what is the expectation. As I mentioned it feels weird to visually see the endpoint field (disabled) with a value when I un-check the box and then have the value disappear after the PUT is complete because the endpoint was actually deleted. The frontend is assigning `null` to the endpoint if the checkbox is `false`. I don't think this is bad if we also empty the endpoint field when you un-check the box. But as it works now it feels odd—or off. The UI reads as I can have an endpoint saved but disable whether I want it used but then the endpoint disappears breaking this expectation. Here's a quick video of what this looks like:

https://user-images.githubusercontent.com/438711/192864356-d3a70cba-e0fc-42e6-b0bd-0b6d522986bd.mp4

So my question is how should this behave? Because I've tested the backend and you can technically save the `use_custom_tts` and `custom_tts_endpoint` separately. Here the backend updated the binary field to `0` but allowed the endpoint to remain:

Screen Shot 2022-09-28 at 11 39 20 AM

If this is the case then the frontend can stay as is but we just need to always send the endpoint value based on its presence, not based on whether the box is checked for either field. If this is not the case and we always want the endpoint to be empty if its associative binary is `0` then we need to empty the endpoint field when you un-check the box or have some other clear indication that un-checking the box is also going to delete the endpoint value on the backend as well.

Admittedly we handle this maybe a couple of ways currently across the app. Most notably there are some optional fieldsets on the Carrier form in which un-checking the box wipes them out of local state. However on the Account/Application forms we have optional HTTP basic auth for webhooks and un-checking the box there leaves the entered values in tact—e.g. they would post on save and persist on the backend and the checkzone would toggle open to reveal them again. I already described in this thread earlier how this works on the Settings form. Perhaps we need to decide on an official house rule about these optional fields and normalize them all across these implementations?

If it's the case that we normalize to one style there are only two viable options IMO. What we have here on this PR where the field value persists but is then deleted programmatically should not be one of them. So here are the two approaches:

1. Un-checking the box empties the fields that are toggled by the box. The user then has to enter them again if they decide they want to do that and check the box again. This ensures values aren't persisted for un-checked boxes in local state and saved to the backend.
2. Un-checking the box doesn't empty the fields and they are persisted in local state and saved to the backend. In this case for the Checkzones the act of saving ultimately toggles the zone back open because the fields are saved on the backend.

Option 1 currently looks like this:

https://user-images.githubusercontent.com/438711/192866785-9b515326-35a4-4d3d-8c89-90593de14061.mp4

Option 2 currently looks like this:

https://user-images.githubusercontent.com/438711/192866407-0565f8af-c16c-487e-8ba2-ca24e1790469.mp4

Personally I prefer option 2 because as we see in the case of the Carrier form I already have Outbound values saved on the backend and the simple act of toggling that field wipes them in local state. If I were to then save the form they would get deleted on the backend. Potentially this is not desirable TBH. So for this stuff I am leaning towards always go with what the user entered/did. If, as a user, I really don't want those fields anymore I can delete the values and save. But simply toggling the fields hidden or shown seems alarming to just automagically lose everything.

@davehorton @xquanluu Thoughts?

davehorton: I agree that option 2 is a bit nicer, as it lets you "recall" the text entry if you uncheck and then re-check the checkbox.

And just to be clear -- is there anything I can/should do on the backend at this point?

kitajchuk: @davehorton I don’t think so. I can create an issue to backlog our decision here about normalizing optional fields UI/UX 👍.

I would also say we could add the ability to update the region to this PR as well since you made that change in the backend too! @xquanluu you can simply not disable the region fields on this PR so that users can update it 👍.

xquanluu: love the discussion, I will update it as same as option 2 video. Actually from 1st commit I follow this approach from Carrier page, but then there is another comment that we should use normal checkbox instead of check zone. 

But it's good for me to see and learn from 2 approaches.

xquanluu: I also really want to have a nice animation of Checkzone here in the PR but as discussed we cannot use CheckZone in this case, so I continue with simple text, check box

I have updated:
1/ enable update Region
2/ update the UI logic follow option 2.

kitajchuk:
@davehorton @xquanluu What’s the state of this PR?

davehorton:
Yes from my standpoint it looks good, and I tested it.  @kitajchuk if it also looks good to you go ahead and merge

davehorton:
Though I guess it needs to be rebased to latest main?

Issue #119, opened 2022-09-27
Feature: Support custom endpoints for Azure speech


davehorton:
As background, currently when we collect speech credentials for Microsoft/Azure we collect a region and a subscription key.  On the back end, we use this information by connecting to an Azure URL/endpoint for that region when we want to use the credentials to do TTS or STT.

Azure also has the ability for a customer to create a [custom neural voice ](https://learn.microsoft.com/en-us/azure/cognitive-services/speech-service/custom-neural-voice) for TTS, and/or to create a [custom speech model](https://learn.microsoft.com/en-us/azure/cognitive-services/speech-service/custom-speech-overview) for STT, and we would like to support these.

To do so, the Add/Edit speech credentials panel should be enhanced in the following way:

When vendor is 'Microsoft', add the following new UI controls to the bottom of the panel (above the Cancel/Save buttons)

- a checkbox labeled: "Use a custom voice"
- an input field labeled: Custom voice endpoint

- a checkbox labeled "Use a custom speech model"
- an input field labeled: Custom speech endpoint

Note that the input fields should only be enabled when the corresponding checkbox is checked.

These fields should then be provided in the POST/PUT to the back end with the following property names in the JSON payload of the request: use_custom_tts (0 or 1), custom_tts_endpoint, use_custom_stt (0 or 1), custom_stt_endpoint.

For backwards compatibility, the webapp should be prepared to receive a payload from a GET which does not contain any of these properties, and in that case the checkboxes should be left unchecked and the input field empty.

This feature needs to be done on the legacy branch (first) and then the main branch.

davehorton:
there is a required change in jambonz-api-server for this.  I will make this now

davehorton:
this has been committed on the latest main branch of jambonz-api-server, so test with that.  For initial testing, you can enter any thing in the custom endpoint field, just make sure it is saved and retrieved properly.

xquanluu:
Start working for legacy branch

xquanluu:
Start working for main branch

kitajchuk:
@xquanluu Here's some notes for when you get to implementing this on the new app.

[...]

PR #122, opened 2022-09-27
Feature #119 Support custom endpoints for Azure speech

kitajchuk:


src/containers/internal/views/speech-services/form.tsx
```diff
@@ -157,6 +166,26 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
         setTtsCheck(false);
       }
 
+      if (credential.data.use_custom_stt) {
```
xquanluu, line 169: May be shorthand

setCustomSttCheck(credential.data.use_custom_stt > 0 ? true : false)

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -157,6 +166,26 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
...
+        setCustomSttCheck(false);
+      }
+
+      if (credential.data.use_custom_tts) {
+        setCustomTtsCheck(true);
+      } else {
+        setCustomTtsCheck(false);
+      }
+
+      if (credential.data.custom_stt_endpoint) {
```
xquanluu, line 181: This will have issue that when refetch happens, and the API is wrong with no update that custom_tts_endpoint is null, then CustomSttEndpoint state will show previous input value from user.

May be in short:
setCustomSttEndpoint(credential.data.custom_stt_endpoint || "")

src/containers/internal/views/speech-services/form.tsx
kitajchuk: Fair point. We can probably short hand the syntax for initially setting Boolean state vars in a few other places in the app. Might be a nice little cleanup to do that 😅

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -93,6 +98,10 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
         use_for_stt: sttCheck ? 1 : 0,
         aws_region: vendor === VENDOR_AWS ? region : null,
         region: vendor === VENDOR_MICROSOFT ? region : null,
+        use_custom_stt: customSttCheck ? 1 : 0,
```
xquanluu, line 101: I think these parameters should be sent only when vendor === VENDOR_MICROSOFT

meybe

...(vendor === VENDOR_MICROSOFT && {
region: region || null,
use_custom_stt: customSttCheck ? 1 : 0,
use_custom_tts: customTtsCheck ? 1 : 0,
custom_stt_endpoint: customSttCheck ? customSttEndpoint || null : null,
custom_tts_endpoint: customTtsCheck ? customTtsEndpoint || null : null,
})

kitajchuk: Good catch 👍

src/containers/internal/views/speech-services/form.tsx
kitajchuk: This was fixed on the backend by @davehorton so this should work fine as is 🤔. Unless you’re trying to say if it returns null on refetch it would still render the previous user value?

PR #125, opened 2022-09-29
feat: add notfound unit test

xquanluu:


xquanluu:
Hi @kitajchuk Can you please help to check this unit test for NotFound. Thanks a lot :)

src/containers/notfound.cy.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import { Routes, Route, BrowserRouter } from "react-router-dom";
+import { AuthContext, AuthStateContext } from "src/router/auth";
+import NotFound from "./notfound";
+
+const NotFoundTestWrapper = (props: Partial<AuthStateContext>) => {
+  return (
+    <AuthContext.Provider value={props as AuthStateContext}>
```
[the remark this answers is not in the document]
kitajchuk: I used this `TestProvider` to create a test for the `RequireAuth` component here:

https://github.com/jambonz/jambonz-webapp/blob/main/src/components/require-auth.cy.tsx#L1-L38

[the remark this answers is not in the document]
xquanluu: Thanks for the hint, start using TestProvider

src/containers/notfound.cy.tsx
```diff
@@ -0,0 +1,32 @@
...
+  );
+};
+
+describe("<NotFound>", () => {
+  it("mounts", () => {
+    cy.mount(<NotFoundTestWrapper authorized={true} />);
+  });
+
+  it("authorized has h1, no login button", () => {
+    cy.mount(<NotFoundTestWrapper authorized={true} />);
```
kitajchuk, line 20: This can just be passed as a Boolean prop like `<NotFoundTestWrapper authorized />`

src/containers/notfound.cy.tsx
```diff
@@ -0,0 +1,32 @@
...
+    cy.get("H1").should("have.text", "That page doesn't exist.");
+    cy.get("a").should("not.exist");
+  });
+
+  it("not authorized has h1 and login buttons", () => {
+    cy.mount(<NotFoundTestWrapper authorized={false} />);
```
kitajchuk, line 26: False is the default for the `authorized` prop so it can be omitted when testing unauthorized states.

src/containers/notfound.cy.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    cy.get("H1").should("have.text", "That page doesn't exist.");
+    cy.get("a").should("have.text", "Log in");
```
kitajchuk, line 29: I don’t think it’s too meaningful in these type of component tests to test the text or className on the button. I would test the things that result from the internal api of the component so just that the button exists and the container can be queried with the `notfound—login` className. Above you could check the opposite basically. And for the h1 you could just check that it exists. The text is static in this component so it’s not really a value add to validate what it renders.

kitajchuk (changes requested):

src/containers/notfound.cy.tsx
xquanluu: thanks for the hint, start using it

src/containers/notfound.cy.tsx
xquanluu: thanks for the hint, start using it

src/containers/notfound.cy.tsx
xquanluu: thanks for the hint, I changed the assertions

src/containers/notfound.cy.tsx
```diff
@@ -13,20 +13,19 @@ const NotFoundTestWrapper = (props: Partial<AuthStateContext>) => {
...
 
   it("not authorized has h1 and login buttons", () => {
-    cy.mount(<NotFoundTestWrapper authorized={false} />);
+    cy.mount(<NotFoundTestWrapper />);
 
-    cy.get("H1").should("have.text", "That page doesn't exist.");
-    cy.get("a").should("have.text", "Log in");
-    cy.get("a").should("have.class", "btn btn--hollow btn--white");
+    cy.get("h1.h2").should("exist");
+    cy.get("div.notfound--login").should("exist");
```
kitajchuk, line 29: Last thing we should still check that the button exists or not. I was just saying that we don’t care about it’s className

PR #127, opened 2022-09-30
Test: Login layout container

xquanluu:


src/containers/login/layout.cy.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { Layout } from "./layout";
+
+const LayoutTestWrapper = () => {
```
kitajchuk, line 4: Actually for this test I’m less concerned with the specifics of the elements you have below but more with the structural nature of its purpose. I would simply test that header and footer exist for one. I think you could then actually keep this test wrapper but add the router setup such that you can validate the render of the `Outlet` from react router dom. This would test the real function of this container layout.

kitajchuk (changes requested):

PR #128, opened 2022-09-30
Test: Login container

xquanluu:


xquanluu:
Still try to understanding cy.spy to check the signin function is call when login button click

kitajchuk:
> Still try to understanding cy.spy to check the signin function is call when login button click

There’s a basic example in this test: https://github.com/jambonz/jambonz-webapp/blob/main/src/components/select-filter.cy.tsx

kitajchuk (changes requested):
I think you can use the newer test utils here now instead which include some utility methods for signin success/failure handling. https://github.com/jambonz/jambonz-webapp/blob/main/src/components/select-filter.cy.tsx

Issue #118, opened 2022-09-26
Feature: Test coverage for src/containers


kitajchuk:
Refer to original test configuration issue: https://github.com/jambonz/jambonz-webapp/issues/102

**Checklist:**

- [ ] notfound
- [ ] login/create-password.
- [ ] login/layout
- [ ] login/login
- [ ] internal/api-keys
- [ ] internal/navi
- [ ] internal/layout
- [ ] internal/accounts
- [ ] internal/alerts
- [ ] internal/applications
- [ ] internal/carriers
- [ ] internal/ms-teams-tenants
- [ ] internal/phone-numbers
- [ ] internal/recent-calls
- [ ] internal/settings
- [ ] internal/speech-services

xquanluu:
start working on  login/create-password

xquanluu:
start login/layout

Issue #117, opened 2022-09-26
Feature: Test coverage for src/components


kitajchuk:
Refer to original test configuration issue: https://github.com/jambonz/jambonz-webapp/issues/102

**Checklist:**

- [ ] access-control
- [ ] account-filter
- [ ] application-filter
- [ ] clipboard
- [ ] icons.ts
- [ ] index.ts
- [ ] modal
- [ ] obscure
- [ ] pagination
- [ ] require-auth
- [ ] search-filter
- [ ] section
- [ ] select-filter
- [ ] spinner
- [ ] toast
- [ ] tooltip
- [ ] forms/account-select
- [ ] forms/application-select
- [ ] forms/checkzone
- [ ] forms/file-upload
- [ ] forms/local-limits
- [ ] forms/message
- [ ] forms/passwd
- [ ] forms/selector

Issue #102, opened 2022-09-16
Feature: Configure cypress component testing


kitajchuk:
Here’s some links. Cypress component testing could maybe suffice as a baseline for running basic tests and we can add the step to our GitHub actions `pr-checks`.
 
It would be interesting to figure out how to get a proper e2e testing suite set up. I don’t think it would be generally easy as it would require running the backend stack and the web app frontend for the suite 🤔.

- https://docs.cypress.io/guides/component-testing/writing-your-first-component-test
- https://docs.cypress.io/guides/component-testing/mounting-react

**Short-term:**

- Everything in `src/components` should  have a test.
- Everything in `src/containers` should have a test.

**Long-term:**

- Everything in `src/utils` should have a test.
- Everything in `src/{store|router|api|vendor` should have a test.

PR #138, opened 2022-10-23
Add inputs to Carrier form to support outbound registration

EgleHelms:
This PR adds the new inputs to the Carrier form on the main branch. 
Inputs added:
- register_from_user 
- register_from_domain 
- register_public_ip_in_contact 

image

### DIFF
--- a/src/containers/internal/views/carriers/form.tsx
+++ b/src/containers/internal/views/carriers/form.tsx
@@ -90,6 +90,9 @@ export const CarrierForm = ({
   const [sipPass, setSipPass] = useState("");
   const [sipRealm, setSipRealm] = useState("");
   const [initialRegister, setInitialRegister] = useState(false);
+  const [fromUser, setFromUser] = useState("");
+  const [fromDomain, setFromDomain] = useState("");
+  const [regPublicIpInContact, setRegPublicIpInContact] = useState(false);
 
   const [prefix, setPrefix] = useState("");
   const [initialPrefix, setInitialPrefix] = useState(false);
@@ -146,12 +149,24 @@ export const CarrierForm = ({
       if (obj.register_sip_realm) {
         setSipRealm(obj.register_sip_realm);
       }
+      if (obj.register_from_user) {
+        setFromUser(obj.register_from_user);
+      }
+      if (obj.register_from_domain) {
+        setFromDomain(obj.register_from_domain);
+      }
+      if (obj.register_public_ip_in_contact) {
+        setRegPublicIpInContact(obj.register_public_ip_in_contact);
+      }
 
       if (
         obj.requires_register ||
         obj.register_username ||
         obj.register_password ||
-        obj.register_sip_realm
+        obj.register_sip_realm ||
+        obj.register_from_user ||
+        obj.register_from_domain ||
+        obj.register_public_ip_in_contact
       ) {
         setInitialRegister(true);
       } else {
@@ -456,6 +471,11 @@ export const CarrierForm = ({
         register_username: sipUser.trim() || null,
         register_password: sipPass.trim() || null,
         register_sip_realm: sipRealm.trim() || null,
+        register_from_user: sipRegister && fromUser ? fromUser.trim() : null,
+        register_from_domain:
+          sipRegister && fromDomain ? fromDomain.trim() : null,
+        register_public_ip_in_contact:
+          sipRegister && regPublicIpInContact ? true : false,
         tech_prefix: prefix.trim() || null,
         diversion: diversion.trim() || null,
         is_active: isActive,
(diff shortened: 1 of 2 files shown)


src/containers/internal/views/carriers/form.tsx
```diff
@@ -456,6 +471,11 @@ export const CarrierForm = ({
         register_username: sipUser.trim() || null,
         register_password: sipPass.trim() || null,
         register_sip_realm: sipRealm.trim() || null,
+        register_from_user: sipRegister && fromUser ? fromUser.trim() : null,
+        register_from_domain:
+          sipRegister && fromDomain ? fromDomain.trim() : null,
+        register_public_ip_in_contact:
+          sipRegister && regPublicIpInContact ? true : false,
```
kitajchuk, line 478: I think you can just do `sipRegister && regPublicIpInContact` as this should properly evaluate to true or false and you could drop the ternary.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -737,6 +760,36 @@ export const CarrierForm = ({
                       required={sipRegister}
                       onChange={(e) => setSipRealm(e.target.value)}
                     />
+                    <label htmlFor="fromUser">SIP From User</label>
```
kitajchuk, line 763: The `htmlFor` needs to reference the value of the `id` attribute on the input elements in order for the label to focus the inputs. All three of these need to be updated with the correct value it looks like.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -737,6 +760,36 @@ export const CarrierForm = ({
...
+                    <label htmlFor="sipFromDomain">SIP From Domain</label>
+                    <input
+                      id="from_domain"
+                      name="from_domain"
+                      type="text"
+                      value={fromDomain}
+                      placeholder="Optional: specify host part of SIP From header"
+                      onChange={(e) => setFromDomain(e.target.value)}
+                    />
+                    <label htmlFor="regPublicIpInContact" className="chk">
```
kitajchuk, line 781: This one probably still works because the label wraps the input but we should still update the `htmlFor` with the correct value 👍

kitajchuk (changes requested):
Couple comments/suggestions but generally looks good. Main thing is I’d like to see the labels get the proper values to reference the input ids.

src/containers/internal/views/carriers/form.tsx
EgleHelms: changed

src/containers/internal/views/carriers/form.tsx
EgleHelms: chnaged

src/containers/internal/views/carriers/form.tsx
EgleHelms: changed

kitajchuk (approved):
Looks good 👍.

kitajchuk:
I’ll let @davehorton have a look but this looks good to me. For merge our preference is to use the Squash and merge strategy 👍

davehorton:
yup, will do.  I will just run a quick functional test.

Question you @kitajchuk : should we have @EgleHelms include cypress test cases as part of this PR, from a learning perspective?

kitajchuk:
> yup, will do. I will just run a quick functional test.
> 
> Question for you @kitajchuk : should we have @EgleHelms include cypress test cases as part of this PR, from a learning perspective?

I would say no right now—reason is that we are still in a backlog state with getting test coverage backfilled. So there’s no test for this particular form component and this is our most complex form. I need to get one good form component test completed so we have a model for it since it will require mocking interceptions for the backend api with cypress.

You can refer to my original notes/thoughts regarding the possibility of proper e2e test suite here https://github.com/jambonz/jambonz-webapp/issues/102. It’s doable but getting it running in CI is another story…

I think cypress component testing is a good initial format and we have backlog issues to achieve this. If anything biting off one of the UI component tests is an easier entry because they are so isolated. Issues for those can be found here:

https://github.com/jambonz/jambonz-webapp/issues/117

https://github.com/jambonz/jambonz-webapp/issues/118

EgleHelms:
I can definitely write some Cypress tests. I have used this test runner
before.

davehorton:
Got it, thnks @kitajchuk .  I will do a quick functional test and then merge this one, and we can address tests as part of the backlog issues.

kitajchuk:
> I can definitely write some Cypress tests. I have used this test runner before.
> […](#)

Fair enough but I wouldn’t include it in this PR since the test needs to cover the entire form and that’s out of scope here. We’re using cypress component testing which is newer and still beta I think. Having proper e2e with cypress for the container views would be great but we have no way of running the full stack in CI to test that here. So I would opt for mocking the backend with cypress and using their interception APIs along with component testing. That should be good enough considering this webapp never had any testing at all before 🤔

davehorton (approved):

EgleHelms:
Well, I guess a model test will be helpful if you think this is a bit
complex to test and we can separate this for now. I will gladly write tests
afterwards.

PR #143, opened 2022-10-27
parse jwt in the signin function to obtain values

EgleHelms:
If we return only token in the response from the api server as per this issue: https://github.com/jambonz/jambonz-api-server/pull/75 , we need to parse the values in the signin function.

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -111,7 +111,8 @@ export interface User {
   permissions: UserPermissions[];
 }
 
-export interface UserLogin extends User {
+export interface UserLogin {
+  user_sid: string;
   token: string;
   force_change: boolean;
 }
--- a/src/router/auth.tsx
+++ b/src/router/auth.tsx
@@ -95,10 +95,9 @@ export const useProvideAuth = (): AuthStateContext => {
           if (response.status === StatusCodes.OK) {
             token = response.json.token;
             setToken(token);
-            const { force_change, user_sid } = parseJwt(token);
 
-            if (force_change) {
-              sessionStorage.setItem(SESS_USER_SID, user_sid);
+            if (response.json.force_change) {
+              sessionStorage.setItem(SESS_USER_SID, response.json.user_sid);
               sessionStorage.setItem(SESS_OLD_PASSWORD, password);
               navigate(ROUTE_CREATE_PASSWORD);
             } else {
--- a/src/router/auth.tsx
+++ b/src/router/auth.tsx
@@ -95,9 +95,10 @@ export const useProvideAuth = (): AuthStateContext => {
           if (response.status === StatusCodes.OK) {
             token = response.json.token;
             setToken(token);
+            const { force_change, user_sid } = parseJwt(token);
 
-            if (response.json.force_change) {
-              sessionStorage.setItem(SESS_USER_SID, response.json.user_sid);
+            if (force_change) {
+              sessionStorage.setItem(SESS_USER_SID, user_sid);
               sessionStorage.setItem(SESS_OLD_PASSWORD, password);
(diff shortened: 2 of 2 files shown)


davehorton (dismissed):

davehorton:
currently, the user_sid is one of the properties encoded in the jwt.

Does that suffice?

kitajchuk:
> currently, the user_sid is one of the properties encoded in the jwt.
> 
> Does that suffice?

Hmm. I don’t know that I like this setup much. The JWT should only be for user scope, permission and sid. The notion of force_change should not be in the JWT. Personally I think the login post should return token, sid and force_change. The usage of decoding the token for the other user data is decoupled from login so it’s generally more usable.

davehorton:
ok, np, we can make that change in jambonz-api-server

kitajchuk:
> ok, np, we can make that change in jambonz-api-server

Okay cool. @EgleHelms once that’s done you can revert the change here since we don’t need to decode JWT in the login flow. But you will need to update some typings and test utils changes made in my PR yesterday since we’re tweaking the dials here. Namely UserLogin no longer needs to extend the user interface so it can just have its own 3 fields etc. You can see the changes I made yesterday here and it should be clear how to change things as needed: https://github.com/jambonz/jambonz-webapp/pull/140/files

davehorton:
ok I pushed [this commit](https://github.com/jambonz/jambonz-api-server/commit/8d303390b7c3e003ad847dd44f4467402116196c) on the main jambonz-api-server branch.

@EgleHelms now you will need to revert back to getting force_change from the json body of the /login response, as it will not be in the jwt.

EgleHelms:
@kitajchuk is this what you had in mind? I reversed the changes and changed the typings. Ran through the tests and they work fine. Tested manually also nothing seems to have any issues logging in. The "add user" block is hidden. Anything else, I should check?

davehorton (approved):

kitajchuk:
> @kitajchuk is this what you had in mind? I reversed the changes and changed the typings. Ran through the tests and they work fine. Tested manually also nothing seems to have any issues logging in. The "add user" block is hidden. Anything else, I should check?

Lemme pull it and check later this morning. I would like to give the api changes an actual runtime check as well 👍

src/api/types.ts
```diff
@@ -111,7 +111,8 @@ export interface User {
   permissions: UserPermissions[];
 }
 
-export interface UserLogin extends User {
+export interface UserLogin {
```
kitajchuk, line 114: Looks like the backend is still returning a full payload. Based on my comment I figured we'd switch the backend to only return `{ token, user_sid, force_change }` for the login post response but it's still also returning `{ scope, permissions }` as well. So I think we need to keep this as it was using the interface extension.

```ts
export interface User {
  scope: UserScopes;
  user_sid: string;
  permissions: UserPermissions[];
}

export interface UserLogin extends User {
  token: string;
  force_change: boolean;
}
```

Based on latest from api server on `main` branch:

Screen Shot 2022-10-28 at 7 44 50 AM

kitajchuk (changes requested):
Just the changes mentioned. Sorry for the back and forth here. I expected something different from the backend changes but it's also fine the way it is 🤷‍♂️

src/api/types.ts
EgleHelms: @davehorton should I adjust the backend? I saw you added the spread of the object back, instead of just the required values. Here: https://github.com/jambonz/jambonz-api-server/commit/8d303390b7c3e003ad847dd44f4467402116196c

kitajchuk: I think it really doesn’t matter all that much at this point. I actually think this PR could just be closed ya?

PR #144, opened 2022-10-28
Add nuance speech provider

EgleHelms:
Added to:
1. Speech provider with credentials clientId and secret (the post request works)
2. Speech Synthesis Vendor with sample mapping of the voices (need your input @davehorton)
3. Speech Recognizer Vendor (I was not sure about this, please let me know)

Question for No2 above, what values do you need to be saved, @davehorton ? Currently, this is the mapping I used:
```
  {
    code: "en-US",
    name: "English (United States)",
    voices: [
      { value: "allison", name: "Allison (standard)", model: "standard" },
      { value: "ava-ml", name: "Ava-Ml (enhanced)", model: "enhanced" },
    ],
  },
```
  
or should we have voices[0].value" filled with the object like this:
```
{
	"language": "ar-ww",
	"name": "Laila",
	"model": "standard",
	"gender": "female"
}

```
Depends on what you need here.

### DIFF
--- a/src/containers/internal/views/speech-services/form.tsx
+++ b/src/containers/internal/views/speech-services/form.tsx
@@ -22,6 +22,7 @@ import {
   VENDOR_AWS,
   VENDOR_GOOGLE,
   VENDOR_MICROSOFT,
+  VENDOR_NUANCE,
   VENDOR_WELLSAID,
 } from "src/vendor";
 import { MSG_REQUIRED_FIELDS } from "src/constants";
@@ -51,6 +52,8 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
   const [apiKey, setApiKey] = useState("");
   const [accessKeyId, setAccessKeyId] = useState("");
   const [secretAccessKey, setSecretAccessKey] = useState("");
+  const [clientId, setClientId] = useState("");
+  const [secretKey, setSecretKey] = useState("");
   const [googleServiceKey, setGoogleServiceKey] =
     useState<GoogleServiceKey | null>(null);
   const [useCustomTts, setUseCustomTts] = useState(false);
@@ -137,6 +140,8 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
             vendor === VENDOR_MICROSOFT || vendor === VENDOR_WELLSAID
               ? apiKey
               : null,
+          client_id: vendor === VENDOR_NUANCE ? clientId : null,
+          secret: vendor === VENDOR_NUANCE ? secretKey : null,
         })
           .then(({ json }) => {
             toastSuccess("Speech credential created successfully");
@@ -193,6 +198,14 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
 
       if (credential.data.aws_region) {
         setRegion(credential.data.aws_region);
+      }
+
+      if (credential.data.client_id) {
+        setClientId(credential.data.client_id);
+      }
+
+      if (credential.data.secret) {
+        setSecretKey(credential.data.secret);
       }
 
       setUseCustomTts(credential.data.use_custom_tts > 0 ? true : false);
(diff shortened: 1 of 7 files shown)


davehorton:
The first mapping you show above looks good.  Let me pull and test this branch and will give more feedback shortly

davehorton:
The way you are showing it in the Applications dropdown looks fine, but under Speech Synthesis I only see choices for Arabic, Basque and English at the moment.  I presume that is because you are waiting for feedback before adding the rest?

However, I realize there is something special that we will need to do for Nuance when it comes to showing voices for speech synthesis and saving them to the server.

On the server, in the Applications table we store only a language and a voice (along with vendor) for the speech synthesis (tts) choice.  For instance
```
speech_synthesis_language: "ar-WW",
speech_synthesis_vendor:  "nuance"
speech_synthesis_voice: "tarik"
```

The problem is, with nuance it is possible that they can offer same {language, voice} in both "standard" and "enhanced" model.  So somehow we need to store the model in the DB along with the voice.

I would suggest that when you POST or PUT an update to /Applications you combine the selected voice & model and submit it like this in the payload

```
speech_synthesis_voice:  "miriam - standard"
```
Accordingly, when retrieving the tts voice for an application from the db you would need to parse the model from the string and set it based on that, as far as selecting the proper item in the dropdown to select.

Does that make sense?

davehorton:
Also, there is a problem with saving a speech credential.  When we save it, we also test it and currently since you dont have a valid client_id and secret this is going to fail.  Yet see below, the UI incorrectly says the credential is valid
image

EgleHelms:
Understood, take a look at the things you tested above.

Is Nuance also a Speech Recognizer? Or it is only AWS, Google and MS and Nuance should not be added there?

davehorton:
Yes, Nuance is also a recognizer so it should offer both stt and tts options (as you currently show it)

davehorton:
also @EgleHelms  can you pull the latest from the feature/nuance-speech for jambonz-api-server for your local testing?  I pushed some fixes there for returning better error info when the tts test fails

EgleHelms:
so I think I could fix the status check easily, this was just a missed check.

davehorton:
ok, let me know when all of the tts languages are in, I will retest then

EgleHelms:
I added all the languages to the map, but I am not sure if I will find a way to parse the "allison - standard" voice value today.  Since I am not totally sure where we need this exactly. The parsing should probably be in the place we are sending the data about the voice selected to the vendor. Somehow it seems like something for the backend.

davehorton:
> @davehorton I added all the languages to the map, but I am not sure if I will find a way to parse the "allison - standard" voice value today. Since I am not totally sure where we need this exactly. The parsing should probably be in the place we are sending the data about the voice selected to the vendor. Somehow it seems like something for the backend.

Good point.  Let me see what I can do on the back end.

davehorton:
@kitajchuk I would like to merge this PR, could you take a quick look and give a review?

davehorton (dismissed):

kitajchuk:
> @kitajchuk I would like to merge this PR, could you take a quick look and give a review?

Yes, sorry 😅

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -304,6 +317,35 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
...
+              id="nuance_client_id"
+              required
+              type="text"
+              name="nuance_client_id"
+              placeholder="Client ID"
+              value={clientId}
+              onChange={(e) => setClientId(e.target.value)}
+              disabled={credential ? true : false}
+            />
+            <label htmlFor="aws_secret_key">
```
kitajchuk, line 335: This `htmlFor` should be for `nuance_secret`

kitajchuk (changes requested):
Looks good! Just the one change so we don’t have buggy label focus 👍

PR #147, opened 2022-11-07
Prefill predefined carrier gateways in a form

EgleHelms:
Added a post request to get the sid for the template and then navigate to the template.
This way we navigate to a template created and to `/internal/carriers/<template_sid>/edit`

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -194,7 +194,10 @@ export const getFetch = <Type>(url: string) => {
   });
 };
 
-export const postFetch = <Type, Payload>(url: string, payload: Payload) => {
+export const postFetch = <Type, Payload>(
+  url: string,
+  payload: Payload | undefined
+) => {
   return fetchTransport<Type>(url, {
     method: "POST",
     ...(payload && { body: JSON.stringify(payload) }),
@@ -286,12 +289,11 @@ export const postCarrier = (sid: string, payload: Partial<Carrier>) => {
 
 export const postPredefinedCarrierTemplate = (
   currentServiceProviderSid: string,
-  predefinedCarrierSid: string,
-  payload: void
+  predefinedCarrierSid: string
 ) => {
-  return postFetch<SidResponse, void>(
+  return postFetch<SidResponse, undefined>(
     `${API_BASE_URL}/ServiceProviders/${currentServiceProviderSid}/PredefinedCarriers/${predefinedCarrierSid}`,
-    payload
+    undefined
   );
 };
 
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -194,9 +194,9 @@ export const getFetch = <Type>(url: string) => {
   });
 };
 
-export const postFetch = <Type, Payload>(
+export const postFetch = <Type, Payload = undefined>(
   url: string,
-  payload: Payload | undefined
+  payload?: Payload
 ) => {
   return fetchTransport<Type>(url, {
     method: "POST",
@@ -291,9 +291,8 @@ export const postPredefinedCarrierTemplate = (
   currentServiceProviderSid: string,
   predefinedCarrierSid: string
 ) => {
-  return postFetch<SidResponse, undefined>(
-    `${API_BASE_URL}/ServiceProviders/${currentServiceProviderSid}/PredefinedCarriers/${predefinedCarrierSid}`,
-    undefined
+  return postFetch<SidResponse>(
+    `${API_BASE_URL}/ServiceProviders/${currentServiceProviderSid}/PredefinedCarriers/${predefinedCarrierSid}`
   );
 };
 
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -197,7 +197,7 @@ export const getFetch = <Type>(url: string) => {
 export const postFetch = <Type, Payload>(url: string, payload: Payload) => {
   return fetchTransport<Type>(url, {
     method: "POST",
-    body: JSON.stringify(payload),
+    body: payload ? JSON.stringify(payload) : null,
     headers: getAuthHeaders(),
   });
 };
@@ -285,10 +285,14 @@ export const postCarrier = (sid: string, payload: Partial<Carrier>) => {
 };
 
 export const postPredefinedCarrierTemplate = (
-  apiPath: string,
+  currentServiceProviderSid: string,
+  predefinedCarrierSid: string,
   payload: void
 ) => {
-  return postFetch<SidResponse, void>(`${API_BASE_URL}/${apiPath}`, payload);
+  return postFetch<SidResponse, void>(
+    `${API_BASE_URL}/ServiceProviders/${currentServiceProviderSid}/PredefinedCarriers/${predefinedCarrierSid}`,
+    payload
+  );
 };
 
 export const postSipGateway = (payload: Partial<SipGateway>) => {
@@ -539,6 +543,7 @@ export const useApiData: UseApiData = <Type>(apiPath: string) => {
 
   useEffect(() => {
     let ignore = false;
+
     getFetch<Type>(`${API_BASE_URL}/${apiPath}`)
       .then(({ json }) => {
         if (!ignore) {
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -197,7 +197,7 @@ export const getFetch = <Type>(url: string) => {
 export const postFetch = <Type, Payload>(url: string, payload: Payload) => {
   return fetchTransport<Type>(url, {
     method: "POST",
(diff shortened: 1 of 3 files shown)


EgleHelms:
@davehorton nope, sorry it is not ready yet. I will tag you when it is.

EgleHelms:
@davehorton @kitajchuk this is ready for review. Please take a look and test. Thank you.

davehorton:
The function of selecting a predefined carrier now works.  However, I see these errors in the console when I log in.  Do you see these as well?

image

davehorton:
Yes, this works in my testing.

davehorton (approved):

davehorton:
@kitajchuk can you give a review when you get a chance?

src/api/index.ts
```diff
@@ -532,7 +539,6 @@ export const useApiData: UseApiData = <Type>(apiPath: string) => {
 
   useEffect(() => {
     let ignore = false;
-
```
kitajchuk, line 535: This line doesn’t need to be removed.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -523,9 +524,23 @@ export const CarrierForm = ({
 
   useEffect(() => {
     if (predefinedName && hasLength(predefinedCarriers)) {
-      setCarrierStates(
-        predefinedCarriers.filter((a) => a.name === predefinedName)[0]
-      );
+      const currentServiceProviderSid =
+        currentServiceProvider?.service_provider_sid;
+      const predefinedCarrierSid =
+        predefinedName.length !== 0
```
kitajchuk, line 530: You should just be able to check if predefinedName is truthy like before I would suspect? It’s a string so no it should be empty or some value.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -523,9 +524,23 @@ export const CarrierForm = ({
 
   useEffect(() => {
     if (predefinedName && hasLength(predefinedCarriers)) {
-      setCarrierStates(
-        predefinedCarriers.filter((a) => a.name === predefinedName)[0]
-      );
+      const currentServiceProviderSid =
```
kitajchuk, line 527: This is a really unnecessary variable. Just reference this where you need it. We should avoid creating variables that are just assigned to some available reference like this.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -523,9 +524,23 @@ export const CarrierForm = ({
...
+        currentServiceProvider?.service_provider_sid;
+      const predefinedCarrierSid =
+        predefinedName.length !== 0
+          ? predefinedCarriers?.filter((a) => a.name === predefinedName)[0]
```
kitajchuk, line 531: Ha should have caught this before but you should just use the find array method here. This is a strange use of filter since really you want find 🤔

src/api/index.ts
```diff
@@ -284,6 +284,13 @@ export const postCarrier = (sid: string, payload: Partial<Carrier>) => {
   );
 };
 
+export const postPredefinedCarrierTemplate = (
+  apiPath: string,
+  payload: void
```
kitajchuk, line 289: I suspect this is okay as it looks like @davehorton has tested and reviewed this already but technically you are sending a JSON stringified body of undefined to the server since postFetch doesn’t currently assume to have no payload…

kitajchuk (changes requested):
I think there’s at least a few reasonable changes to be made here but overall looks good 👍

src/api/index.ts
EgleHelms: @kitajchuk would you suggest just making it a simple post request to not reuse this function?

src/api/index.ts
EgleHelms: fixed.

src/api/index.ts
EgleHelms: @kitajchuk I refactored a bit, let me know what you think.

src/api/index.ts
```diff
@@ -197,7 +197,7 @@ export const getFetch = <Type>(url: string) => {
 export const postFetch = <Type, Payload>(url: string, payload: Payload) => {
   return fetchTransport<Type>(url, {
     method: "POST",
-    body: JSON.stringify(payload),
+    body: payload ? JSON.stringify(payload) : null,
```
kitajchuk, line 200: I think if we want to make a change to this core function like this it would be better to omit the body entirely if it is undefined as an argument here. More like `…(payload && { body: JSON.stringify(payload) })`

EgleHelms: Yep, that works well. Thanks.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -523,9 +524,23 @@ export const CarrierForm = ({
 
   useEffect(() => {
     if (predefinedName && hasLength(predefinedCarriers)) {
-      setCarrierStates(
-        predefinedCarriers.filter((a) => a.name === predefinedName)[0]
-      );
+      const predefinedCarrierSid = predefinedName
```
kitajchuk, line 527: Since predefinedName is known to be set five the wrapping condition I don’t think you need a ternary here. Just call the find.

src/api/index.ts
```diff
@@ -197,7 +197,7 @@ export const getFetch = <Type>(url: string) => {
 export const postFetch = <Type, Payload>(url: string, payload: Payload) => {
   return fetchTransport<Type>(url, {
     method: "POST",
-    body: JSON.stringify(payload),
+    ...(payload && { body: JSON.stringify(payload) }),
```
kitajchuk, line 200: I think we need to make the generic here something like `Payload | undefined` don’t you think 🤔? Reasoning is we’re now saying this is a valid option for this function. Also will have a note on use of void below…

src/api/index.ts
```diff
@@ -284,6 +284,17 @@ export const postCarrier = (sid: string, payload: Partial<Carrier>) => {
   );
 };
 
+export const postPredefinedCarrierTemplate = (
+  currentServiceProviderSid: string,
+  predefinedCarrierSid: string,
+  payload: void
```
kitajchuk, line 290: This is not a proper use case for the void type. We should accept that payload can be undefined in the postFetch method and you can omit payload here 👍

src/api/index.ts
EgleHelms: I added the undefined. I am actually surprised that typescript didn't note me on this.

src/api/index.ts
```diff
@@ -194,10 +194,13 @@ export const getFetch = <Type>(url: string) => {
   });
 };
 
-export const postFetch = <Type, Payload>(url: string, payload: Payload) => {
+export const postFetch = <Type, Payload>(
```
kitajchuk, line 197: I believe my intention was you would make the generic here accept undefined as well and then you can omit passing it at all below.

src/api/index.ts
```diff
@@ -284,6 +287,16 @@ export const postCarrier = (sid: string, payload: Partial<Carrier>) => {
   );
 };
 
+export const postPredefinedCarrierTemplate = (
+  currentServiceProviderSid: string,
+  predefinedCarrierSid: string
+) => {
+  return postFetch<SidResponse, undefined>(
+    `${API_BASE_URL}/ServiceProviders/${currentServiceProviderSid}/PredefinedCarriers/${predefinedCarrierSid}`,
+    undefined
```
kitajchuk, line 296: If you modify the generic for postFetch you can omit any payload code here since undefined would be the default of not passing a type or argument 🤔

src/api/index.ts
EgleHelms: @kitajchuk can you let me know the above?

kitajchuk: To make payload optional I think you can do `<Type, Payload = void>`

src/api/index.ts
davehorton: @EgleHelms  is this fix still outstanding?  I'd like to see this PR get merged..

src/api/index.ts
EgleHelms: @kitajchuk the postFetch still asks for two arguments when referred to in postPredefinedCarrierTemplate

EgleHelms: I think I figured it out.

src/api/index.ts
EgleHelms: I will ask Brandon for another rereview. Struggled with types a bit.

kitajchuk (approved):

EgleHelms:
@davehorton I think this now can be merged.

PR #149, opened 2022-11-11
Update Node Image

guilherme-rauen:
This PR simplifies the Docker build commands and updates the Node image to the latest stable version. And it is the most secure one, with no vulnerabilities on Snyk scan.


kitajchuk:
@davehorton these changes would be good for the main branch as well I think.

PR #150, opened 2022-11-11
Feature/users dashboard

EgleHelms:
Adding User List view on the Nav
Added filtering mechanism to the List view, currently only the search filter is working.
Added user form to implement basic add/edit/delete functionality

Right now only admin users can be created with the UI.


src/containers/internal/navi/items.ts
```diff
@@ -22,6 +23,11 @@ export interface NaviItem {
 }
 
 export const naviTop: NaviItem[] = [
+  {
+    label: "Users",
+    icon: Icons.Users,
```
kitajchuk, line 28: If we’re going to use this icon we should find a different one for ms teams 🤔

src/containers/internal/views/users/delete.tsx
```diff
@@ -0,0 +1,28 @@
...
+  handleSubmit: () => void;
+};
+
+export const DeleteUser = ({
+  user,
+  handleCancel,
+  handleSubmit,
+}: DeleteProps) => {
+  return (
+    <>
```
kitajchuk, line 18: This fragment is not necessary

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+  };
+
+  const handleDelete = () => {
+    if (user) {
+      deleteUser(user.user_sid)
+        .then(() => {
+          navigate(ROUTE_INTERNAL_USERS);
+          toastSuccess(
+            <>
+              Deleted account <strong>{user.name}</strong>
```
kitajchuk, line 40: Should be deleted user

src/components/service-provider-filter.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { useEffect, useState } from "react";
```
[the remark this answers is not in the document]
davehorton: eventually, we would want to be able to select by
- admin level users
- service provider level users for a given service provider
- service provider level users for a given service provider and users of accounts underneath that SP
- account level users for a given account (and by implication, service provider)
- all users

I think right now since the current focus was adding support for admin-level users we have just not implemented all of this

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+      setScope("admin");
+    } else if (user.service_provider_sid) {
+      setScope("service_provider");
+    } else {
+      setScope("account");
+    }
+    return;
+  };
+
+  const isValidPasswd = (password: string) => {
```
kitajchuk, line 61: I would prefer this is moved into untils and you pass the fetched settings as a second argument

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+    let result;
+    if (pwdSettings) {
+      result =
```
kitajchuk, line 64: Also you can just return here and return false at the end, no need for the variable

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+          refetch();
+          toastSuccess("User created successfully");
+          navigate(`${ROUTE_INTERNAL_USERS}/${json.user_sid}/edit`);
+        })
+        .catch((error: { msg: IMessage }) => {
+          toastError(error.msg);
+        });
+    }
+
+    if (user && !modal) {
```
kitajchuk, line 100: What context would there be a modal?

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+        });
+    }
+  };
+
+  /** Set current user data values if applicable -- e.g. "edit mode" */
+  useEffect(() => {
+    if (user) {
+      setName(user.name);
+      setForceChange(!!user.force_change);
+      getScope(user);
```
kitajchuk, line 124: I don’t like this idea of a “getter” that sets state. The getter should be in utils and return the value you would set in state here.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+            />
+          </fieldset>
+          <fieldset>
+            <label htmlFor="email">
+              User email<span>*</span>
+            </label>
+            <input
+              id="email"
+              required
+              type="text"
```
kitajchuk, line 182: This should use the email input type

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+              name="email"
+              placeholder="User Email"
+              value={email}
+              onChange={(e) => setEmail(e.target.value)}
+            />
+          </fieldset>
+          <fieldset>
+            <label htmlFor="initialPassword">
```
kitajchuk, line 190: This labels for needs to match the input id

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+                type="checkbox"
+                checked={forceChange}
+                onChange={(e) => setForceChange(e.target.checked)}
+              />
+              <div>Force change of password</div>
+            </label>
+          </fieldset>
+          <fieldset>
+            {!window.location.href.includes("/add") && (
+              <ButtonGroup>
```
kitajchuk, line 216: This button should be in the other button group

src/containers/internal/views/users/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+  AccountFilter,
+  SelectFilter,
+  ServiceProviderFilter,
+} from "src/components";
+// import { DeleteUser } from "./delete";
+import { hasLength, hasValue, useFilteredResults } from "src/utils";
+
+import type { ServiceProvider, Account, User } from "src/api/types";
+
+const scopeSelection = [
```
kitajchuk, line 23: This should be in api constants

src/containers/internal/views/users/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+        <ServiceProviderFilter
+          serviceProvider={[serviceProviderSid, setServiceProviderSid]}
+          serviceProviders={serviceProviders}
+        />
+      </section>
+
+      <Section {...(hasLength(filteredUsers) && { slim: true })}>
+        <div className="grid grid--col4">
+          <div className="grid__row grid__th">
+            <div>Users</div>
```
kitajchuk, line 132: We will want this UI to be more like the others with icons etc. I can help work that out. @davehorton do you think users will be very large data sets like accounts etc? I am wondering if here we could use the card view? But maybe not, you’re call 👍

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+          );
+        })
+        .catch((error) => {
+          toastError(error.msg);
+        });
+    }
+  };
+
+  const getScope = (user: User) => {
+    if (!user.account_sid && !user.service_provider_sid) {
```
kitajchuk, line 51: I would also change the logic here. If account sid return account. Same for service provider and return admin as default if those two are not set. This way seems more clear

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,243 @@
...
+import { toastError, toastSuccess } from "src/store";
+import { deleteUser, postFetch, putUser, useApiData } from "src/api";
+import { ROUTE_INTERNAL_USERS } from "src/router/routes";
+import { ClipBoard, Section } from "src/components";
+import { DeleteUser } from "./delete";
+import { MSG_REQUIRED_FIELDS } from "src/constants";
+import { API_USERS } from "src/api/constants";
+
+import type { UserSidResponse, User, PasswordSettings } from "src/api/types";
+import { IMessage } from "src/store/types";
```
kitajchuk, line 14: You should make this import type

src/containers/internal/api-keys/styles.scss
```diff
@@ -98,4 +98,13 @@
       }
     }
   }
+
+  &--col4 {
```
kitajchuk, line 102: Where are we using this?

src/components/service-provider-filter.tsx
```diff
@@ -0,0 +1,75 @@
...
+  label = "Service Provider",
+  serviceProvider: [serviceProviderSid, setServiceProviderSid],
+  serviceProviders,
+  defaultOption,
+}: ServiceProviderFilterProps) => {
+  const [focus, setFocus] = useState(false);
+  const classes = {
+    smsel: true,
+    "smsel--filter": true,
+    "account-filter": true,
```
kitajchuk, line 28: This should be changed

src/containers/internal/views/users/index.tsx
davehorton: I think it could be as large as accounts, given that on a deployment that pushes customer self-care to the account level we'll see at least one user per account.

I have to say I don't love the card view, at least on the desktop.  Maybe its just that I'm not used to seeing it in other products

src/styles/_filters.scss
```diff
@@ -30,6 +30,17 @@
     }
   }
 
+  &--mix {
```
kitajchuk, line 33: You could remove this and just pair the ender and multi classes 👍

src/styles/index.scss
```diff
@@ -111,6 +111,7 @@ details {
   border: 0;
   padding: 0;
   font-size: 0;
+  color: var(--jambonz);
```
kitajchuk, line 114: This should not be here. Apply the color where you need it contextually. This is to remain a very generic className 👍

src/containers/internal/views/users/index.tsx
kitajchuk: Lol fair enough. I’ll let it die 😅🙃

src/components/service-provider-filter.tsx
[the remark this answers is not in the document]
kitajchuk: Okay that’s helpful and also complex lol. So for now what is this service provider filter doing? Filtering only by service provider sid? Is this what we want for now?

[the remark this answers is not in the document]
davehorton: For now, I think the service provider filter should let you either choose "all" or a specific, single service provider.  (I think it needs to be a dropdown).  I dont think there will ever be a need to filter in a way that selects more than one service provider.

So yes I think this is what we want for now.

If you do select a service provider, then you should definitely see all the users for that service provider in the tableview.  The account dropdown should offer you some choices that _may_ add some other users into the filtered tablesview
- service provider users only (nothing added)
- specific account (showing name in dropdown) - adds in users for the selected accounts
- all accounts - adds in users from all accounts under this service provider

I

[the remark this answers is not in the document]
kitajchuk: Okay that seems fairly straightforward for now. @EgleHelms we have examples of combination filters (search+account) on other list views. We leverage memo for the results so it’s a combo thing. This one is different in that there are two filters being applied on top of the fuzzy search. Give it a shot and just reach out if you have questions or need help 👍

src/containers/internal/api-keys/styles.scss
EgleHelms: I kind of reused the API keys table design to achieve this view for users:
image

src/containers/internal/navi/items.ts
EgleHelms: when do you see this icon that refers to the ms teams? Is it ever part of the navigation?
Do you really think it will be confusing?

src/containers/internal/views/users/form.tsx
EgleHelms: When you would click on "Delete user" a modal would come up and for some reason would run the "handleSubmit", I am actually not sure why it is run. Since onClick only changes the modal value to true and according to that we show the modal. WHen the modal is shown the handleSubmit is automatically run, but the handleSubmit is a prop we pass to the DeleteUser component which is the handleDelete function.

src/containers/internal/views/users/form.tsx
EgleHelms: so with the current code it look like this:
image

If I add it to the existing btn group it look slike this:
image

Let me know if I should leave it this way.

src/containers/internal/views/users/index.tsx
EgleHelms: making no changes here.

src/styles/_filters.scss
EgleHelms: I was trying to reuse the styles already written combined, but it was adding extra space this way. Then I need to make the below code conditional somehow. Since I don't want to mess up the other filters. Seems the "Recent calls" one will be affected.

```
    > :first-child {
      margin-left: auto;
    }
```

src/components/service-provider-filter.tsx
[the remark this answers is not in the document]
EgleHelms: This was my take on the filtering.

src/containers/internal/views/users/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+      if (accountSid)
+        setFilterResult(
+          filterResult.filter((e) => {
+            return accountSid === e.account_sid;
+          })
+        );
+
+      setMaxPageNumber(Math.ceil(users.length / Number(perPageFilter)));
+    }
+  }, [
```
EgleHelms, line 75: @kitajchuk this was my approach to filtering - this is still not working as I expect. Issues this filter still currently has:
- filtering works only if it is applied one after the other.
- after I select the filters, changing the value of the filter again will not filter correctly. You need to select "all" again and you can then filter again
- we should fetch accounts from the api only after receiving the serviceProviderSid, to get them only for that provider. but I am not able to get the values inside the useMemo.

src/containers/internal/navi/items.ts
kitajchuk: It is part of the navigation. To see it you need to add an FQDN for the currently service provider and it’s the last item in the BYO section. Yes we should use different icons for these two navigation items. I can poke at feather icons and see what I think 🤔

src/containers/internal/views/users/form.tsx
kitajchuk: This sounds odd. I’ll look again because it should not be the case that this happens like that 🤔

src/containers/internal/views/users/form.tsx
kitajchuk: I think it should be in the group but the last item and aligned right. Also probably gray vs the dark. We have a delete button on the service provider settings as an example.

src/containers/internal/api-keys/styles.scss
kitajchuk: Oh. Idk about this. It doesn’t feel like any of our other list views. I’d prefer to see it more like that. I’m not sure why SID is important here because it’s not elsewhere? I would drop it and make this feel like our other lists with icons 🤔. But maybe @davehorton  feels differently?

kitajchuk: I see that’s fair enough

src/styles/_filters.scss
kitajchuk: I see that’s fair enough

src/containers/internal/views/users/index.tsx
kitajchuk: There should not be a none for service provider. Most views rely on fetching data based on current service provider. In fact we found a lot of scenarios where that was not happening and fixed it during the rewrite. But not all cases. @davehorton is the source of requirements there. I can pull this branch and actually run code and test to provide better feedback on this probably tomorrow morning.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,234 @@
...
+              <div>Force change of password</div>
+            </label>
+          </fieldset>
+          <fieldset>
+            <ButtonGroup left>
+              <Button small subStyle="grey" as={Link} to={ROUTE_INTERNAL_USERS}>
+                Cancel
+              </Button>
+              {user && (
+                <ButtonGroup>
```
kitajchuk, line 212: This should not be a nested button group. Just put the delete button only in the parent group.

src/containers/internal/views/users/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+      </Section>
+      <Section clean>
+        <Button small as={Link} to={`${ROUTE_INTERNAL_USERS}/add`}>
+          Add user
+        </Button>
+      </Section>
+      <footer>
+        <ButtonGroup>
+          {hasLength(users) && (
+            <Pagination
```
kitajchuk, line 168: Is this actually a paginated result set? If so then this should work quite a bit differently since you are fetching the users flat with an api call. You can see how we do paginated data in recent calls and alerts but that data is paginated on the backend. If this is not actually a paginated response from the backend then a lot of code was copy pasted here that should not be here…

src/styles/_forms.scss
```diff
@@ -168,6 +170,15 @@ fieldset {
   input {
     margin: 0 ui-vars.$px01 0 0;
   }
+
+  &--filter {
```
kitajchuk, line 174: What is this for?

src/utils/index.ts
```diff
@@ -133,3 +143,13 @@ export {
   useRedirect,
   useFilteredResults,
 };
+
+export const getUserScope = (user: User) => {
```
kitajchuk, line 147: You could specify the return type here so it’s not just inferred as a string 🤔

src/containers/internal/api-keys/styles.scss
EgleHelms: Initially, I made it like the Application/Account list ones and @davehorton asked me to change it to a table-like view so I reused the API keys one. We can change that to be more like that table with a simpler "Edit" button (Similar to "Delete" in API keys) and I can omit the SID to make it more like the API key table so it is more streamlined. Let me know.

src/containers/internal/views/users/index.tsx
EgleHelms: @kitajchuk I would be super grateful. I will still tinker around with the filter today to see if I can still make the filter better on my own.
Regarding the Service Provider selection - I understood as much. 
@davehorton what role do you think the current "service provider" plays in the view of the User list?

src/containers/internal/views/users/index.tsx
EgleHelms: I thought I might be able to reuse it, so I copied it in for reference as something still to do. 
But yeah, this seems to only be designated for query calls. Clear, will remove it.

src/containers/internal/api-keys/styles.scss
davehorton: I dont think the user sid would be required in the table/listview

src/containers/internal/views/users/index.tsx
davehorton: I dont think we can or should list "admins" as a selection in the main service provider dropdown.  So much of the entire UI depends on their being a current, single, service_provider_sid and I dont want to change that.

@EgleHelms I am fine with your idea of removing the service provider filter and filtering by the "current service provider".  In that case, for an admin user, the list view would encompass all admin users + all users of the current service provider + all users of subsidiary accounts.  I suppose then the controls needed would be
- checkbox: whether or not to show admin users
- dropdown: list of accounts, including a selection for "none" or "all".

Do you agree?

EgleHelms: @davehorton I agree with most of the things, but maybe it also makes sense to have the scope listed in the table + be able to select from dropdown like this:
image

This way this filter will be useful in all views of all users. To know what user has what scope at any point

src/containers/internal/views/users/form.tsx
EgleHelms: Would be super helpful.

src/styles/index.scss
```diff
@@ -113,6 +113,12 @@ details {
   font-size: 0;
 }
 
+.btns--spaced {
```
EgleHelms, line 116: I added this to space buttons like this, not sure how to do it otherwise for the ButtonGroup:
image

src/containers/internal/navi/items.ts
kitajchuk: Ya I think we should change the teams icon.

src/containers/internal/api-keys/styles.scss
EgleHelms: @kitajchuk @davehorton but do you both agree on this view? 

image

kitajchuk: You should be able to delete here like the other list views as well. I’ll smoke and poke this branch in a bit and see how this layout handles responsively

EgleHelms: @kitajchuk I had it like this before - for some reason icons aligned in the same way were not actually aligned, but would this be ok, see below?
image

src/containers/internal/navi/items.ts
EgleHelms: Maybe one of these could work for Teams Tenants?
||||
|--|--|--|
|image |image| image|

src/containers/internal/views/users/form.tsx
EgleHelms: I added this check directly to the handleSubmit to avoid this.

kitajchuk: I think you just need to add type=button to the delete button. The default type for a button is submit so my guess is it’s trying to also submit the form. I’d you do that you can remove this check here. We need to be careful with non-submit buttons in the forms 👍

kitajchuk:
@EgleHelms been busy and unable to run this branch yet for a proper review there. Hoping to be able to do that today 🤞

src/components/index.ts
```diff
@@ -13,6 +13,7 @@ import { SelectFilter } from "./select-filter";
 import { Pagination } from "./pagination";
 import { ApplicationFilter } from "./application-filter";
 import { SearchFilter } from "./search-filter";
+import { ServiceProviderFilter } from "./service-provider-filter";
```
kitajchuk, line 16: If we're not using the service provider filter right now than we should trash it please.

src/containers/internal/api-keys/styles.scss
```diff
@@ -98,4 +98,13 @@
...
+    .grid__row {
```
kitajchuk, line 103: This whole grid scss file should now be moved to root styles and imported in the main styles index like other global things. You can remove the import in api-keys then also.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const [isActive, setIsActive] = useState(true);
+  const [forceChange, setForceChange] = useState(true);
+  const [modal, setModal] = useState(false);
+
+  const handleCancel = () => {
+    setModal(false);
+  };
+
+  const handleSelfDetete = () => {
+    const decodedJwt = parseJwt(getToken());
```
kitajchuk, line 45: We should not be parsing the JWT here. That is done when you log in. You should be selecting the user state from the global store.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,244 @@
...
+    if (user?.data?.user_sid === decodedJwt.user_sid) {
+      localStorage.clear();
```
kitajchuk, line 47: You need to be using the `signout` method from our `useAuth` context provider for this. Example is in the root internal layout but looks like this: `const { signout } = useAuth();`. Then just call `signout()` and be done.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,244 @@
...
+          <fieldset>
+            <ButtonGroup left className="btns--spaced">
+              <Button small subStyle="grey" as={Link} to={ROUTE_INTERNAL_USERS}>
+                Cancel
+              </Button>
+              <Button type="submit" small>
+                Save
+              </Button>
+              {user && (
+                <Button small subStyle="grey" onClick={() => setModal(true)}>
```
kitajchuk, line 227: Ya just add `type="button"` to this and you don't need to check the modal state var in the submit handler 👍

src/containers/internal/views/users/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const [accounts] = useServiceProviderData<Account[]>("Accounts");
+
+  const usersFiltered = useMemo(() => {
+    const serviceProviderUsers = users?.filter((e) => {
+      return (
+        e.scope === "admin" ||
+        e.service_provider_sid ===
+          currentServiceProvider?.service_provider_sid ||
+        accounts?.every((account) => {
+          account.account_sid === e.account_sid;
```
kitajchuk, line 37: You need to actually return the value here for `every`. This syntax is not returning your comparison operator since you're using bracket notation for the arrow function. So more like `accounts?.every((account) => account.account_sid === e.account_sid)`.

src/containers/internal/views/users/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+        })
+      );
+    });
+
+    return serviceProviderUsers
+      ? serviceProviderUsers.filter((e) => {
+          return scopeFilter === "all" && !accountSid
```
kitajchuk, line 44: This is hard to read and grock easily. I might be more of a fan of a straight forward function block with clear if syntax and multiple returns for each condition for clarity here.

src/containers/internal/views/users/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+          options={USER_SCOPE_SELECTION}
+        />
+        <AccountFilter
+          account={[accountSid, setAccountSid]}
+          accounts={accounts}
+          defaultOption={true}
+        />
+      </section>
+
+      <Section {...(hasLength(filteredUsers) && { slim: true })}>
```
kitajchuk, line 89: You can remove this conditional prop and just always have the `slim` prop on this view. With the tables we need it like that or you will get double padding on the empty views like this:

Screen Shot 2022-11-16 at 7 48 47 AM

src/containers/internal/views/users/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+        <div className="grid grid--col4">
+          <div className="grid__row grid__th">
+            <div>Users</div>
```
kitajchuk, line 92: This should be `Name` for the column label 🤔

src/containers/login/create-password.tsx
```diff
@@ -42,7 +43,7 @@ export const CreatePassword = () => {
       return;
     }
 
-    if (!isValidPasswd(password)) {
+    if (passwdSettings && !isValidPasswd(password, passwdSettings)) {
```
kitajchuk, line 46: Not that I have a great answer for this but I'm curious what we would do here if `PasswordSettings` response failed and this was never set. Seems edge case but a thought nonetheless 🤔

src/store/actions.ts
```diff
@@ -63,8 +63,9 @@ export const currentServiceProviderAction = (
 
 export const userAsyncAction = async (): Promise<User> => {
   const token = getToken();
-  const { user_sid, permissions, scope } = parseJwt(token);
-  return { user_sid, permissions, scope };
+  const { user_sid, permissions, scope, force_change, is_active, name, email } =
```
kitajchuk, line 66: If these are all the values that are encoded in the JWT then we can just return `parseJwt(token)` and avoid the destructuring nonsense here 👍

src/containers/internal/navi/items.ts
kitajchuk: I think we keep the users icon for teams and do user plus or user check for the users view.

EgleHelms: changed to usersCheck icon.

src/components/index.ts
EgleHelms: trashed

src/containers/internal/api-keys/styles.scss
EgleHelms: moved

src/containers/login/create-password.tsx
EgleHelms: added a constant to api/constants and added it as a fallback value.

src/containers/internal/views/users/index.tsx
EgleHelms: @kitajchuk sorry, not sure what you mean here.

EgleHelms: oh ok, got it... just that the label should be Name not user, since we list names under it.

src/styles/index.scss
```diff
@@ -113,6 +113,12 @@ details {
...
+  > :nth-child(3) {
```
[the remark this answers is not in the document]
EgleHelms: This does not work in this case.
This refers to the Cancel/Save/Delete buttons in both edit and new user forms. In a new user form Delete button does not exist. Thus the last-child is Save and is spaced.

EgleHelms:
@kitajchuk I think I will branch of this branch to continue with the next part of the implementation (creating scope-specific users, hiding elements according to user scope, other small improvements of the current implementation) since this is getting really big PR and I don't think it is ready to merge, since the api-server branch is also not merged yet.

src/styles/index.scss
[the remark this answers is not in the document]
kitajchuk: I would apply this className conditionally then and use last child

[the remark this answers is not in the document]
EgleHelms: done

EgleHelms:
@kitajchuk this is now ready for your re-review.

I have added changes from the PR I have closed for now, since we agreed to finalize this:
- allow creating scoped users in the form
- improve the temporary password field
- remove delete button from the Users list as per agreement with @davehorton 

I will start improving <AccessControl> component and applying it in another branch. If you have some specifications for me, please create an Issue with a detailed description and assigned me there. This way maybe we can reach agreements on things a bit quicker.

EgleHelms:
@kitajchuk @davehorton can you please re-review this? Dave mentioned we can merge it if this is in order.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,320 @@
...
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+
+    if (scope === "account" && !accounts?.length) {
+      toastError("Cannot create an account. Service Provider has no accounts.");
+      return;
+    }
+
+    if (!user) {
+      passwdCheck();
```
kitajchuk, line 102: You need to return true/false in this method and use that to return here. Returning in the called method will not stop this function block from executing which I imagine we want to do if password is invalid 🤔

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,320 @@
...
+          toastSuccess("User created successfully");
+          navigate(`${ROUTE_INTERNAL_USERS}/${json.user_sid}/edit`);
+        })
+        .catch((error: { msg: IMessage }) => {
+          toastError(error.msg);
+        });
+    }
+
+    if (user && user.data) {
+      if (initialPassword) {
```
kitajchuk, line 129: Same you need to actually return in this function block to stop the api call…

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,320 @@
...
+          <fieldset>
+            {currentUser?.scope !== "account" && (
+              <>
+                <label htmlFor="scope">Scope:</label>
+                <Selector
+                  id="scope"
+                  name="scope"
+                  value={scope}
+                  options={
+                    currentUser?.scope === "service_provider"
```
kitajchuk, line 186: I don’t like this because it’s unclear and if the order changes this would fail us. We do something like this in applications form with vendor options but we use array filter so the things we are excluding are explicit.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,320 @@
...
+      }
+
+      putUser(user.data.user_sid, {
+        name: name || user.data.name,
+        email: email || user.data.email,
+        initial_password: initialPassword || null,
+        force_change: forceChange || !!user.data.force_change,
+        is_active: isActive || !!user.data.is_active,
+        service_provider_sid:
+          scope === "admin" && currentUser?.scope === "admin"
```
kitajchuk, line 140: I think at this point I’d like to see these scope strings as constants similar to how we do with vendors. This would mean updating to use said constants anywhere we are referencing these string values.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,320 @@
...
+                    currentUser?.scope === "service_provider"
+                      ? USER_SCOPE_SELECTION.slice(2)
+                      : USER_SCOPE_SELECTION.slice(1)
+                  }
+                  onChange={(e) => setScope(e.target.value as UserScopes)}
+                />
+              </>
+            )}
+
+            {accounts && accounts.length && scope === "account" ? (
```
kitajchuk, line 195: You can replace this first part with our hasLength(accounts) utility method.

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,320 @@
...
+              value={email}
+              onChange={(e) => setEmail(e.target.value)}
+            />
+          </fieldset>
+          <fieldset>
+            <label htmlFor="initial_password">
+              Temporary password
+              {!user && <span>*</span>}
+            </label>
+            <Passwd
```
kitajchuk, line 271: You need to pass id=“ initial_password” to this component for the label to focus it

src/containers/internal/views/users/form.tsx
```diff
@@ -0,0 +1,320 @@
...
+          </fieldset>
+          <fieldset>
+            <ButtonGroup left className={user && "btns--spaced"}>
+              <Button small subStyle="grey" as={Link} to={ROUTE_INTERNAL_USERS}>
+                Cancel
+              </Button>
+              <Button type="submit" small>
+                Save
+              </Button>
+              {user && (
```
kitajchuk, line 297: Probably should be user && user.data && …

src/containers/internal/views/users/form.tsx
kitajchuk: I also don’t think the ternary is necessary since these conditions require opposite regarding scope equal or not to account. You can just make them each their own conditional render line. It’s preferential to avoid ternary in render IMO if possible 🤔

kitajchuk (changes requested):
Okay some more changes to see here but close 👍

EgleHelms:
@kitajchuk thanks. I am learning so much :) This is again ready for re-review.

src/containers/internal/views/users/form.tsx
```diff
@@ -174,47 +178,48 @@ export const UserForm = ({ user }: UserFormProps) => {
...
+                  <label htmlFor="account">
+                    Account:<span>*</span>
+                  </label>
+                  <input
+                    id="account"
+                    required
+                    type="text"
+                    disabled
+                    name="account"
+                    value={"No accounts."}
```
kitajchuk, line 217: Heh, no need for brackets around string props

kitajchuk (approved):
I’m approving this but can you please address the couple simple comments before merging 🙏

kitajchuk:
Also @davehorton are we intending to merge this to main?

davehorton:
> Also @davehorton are we intending to merge this to main?

Yes

davehorton (dismissed):

kitajchuk:
> > Also @davehorton are we intending to merge this to main?
> 
> Yes

Sweet smash the button

PR #151, opened 2022-11-14
Add deepgram as STT

EgleHelms:
- Added "deepgram" as STT selection in Speech services of an Application (map of languages to be discussed
- Added "deepgram" into the speech-services credentials form with an api key. Disabled the TTS checkbox
- fixed hiding STT checkbox for "Wellsaid", since it is only a synthesis tool

### DIFF
--- a/src/containers/internal/views/speech-services/form.tsx
+++ b/src/containers/internal/views/speech-services/form.tsx
@@ -24,6 +24,7 @@ import {
   VENDOR_MICROSOFT,
   VENDOR_NUANCE,
   VENDOR_WELLSAID,
+  VENDOR_DEEPGRAM,
 } from "src/vendor";
 import { MSG_REQUIRED_FIELDS } from "src/constants";
 import { getObscuredSecret } from "src/utils";
@@ -137,7 +138,9 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
           access_key_id: vendor === VENDOR_AWS ? accessKeyId : null,
           secret_access_key: vendor === VENDOR_AWS ? secretAccessKey : null,
           api_key:
-            vendor === VENDOR_MICROSOFT || vendor === VENDOR_WELLSAID
+            vendor === VENDOR_MICROSOFT ||
+            vendor === VENDOR_WELLSAID ||
+            vendor === VENDOR_DEEPGRAM
               ? apiKey
               : null,
           client_id: vendor === VENDOR_NUANCE ? clientId : null,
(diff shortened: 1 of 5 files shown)


src/vendor/speech-recognizer/deepgram-speech-recognizer-lang.ts
```diff
@@ -0,0 +1,48 @@
...
+    tier: "enhanced",
+  },
+  {
+    name: "English - general - base",
+    code: "en",
+    model: "general",
+    tier: "base",
+  },
+  {
+    name: "English - general - enhanced",
```
EgleHelms, line 41: @davehorton how should I map the language vs list them in the dropdown. 
Here is the language list:https://developers.deepgram.com/documentation/features/language/ 
They seem to have a "model" and "tier". 

we also use "code" as a key for the index, so they have to be unique and when the tier or model changes, the code stays the same, which makes it problematic. Let me know your thoughts on how we should approach this, since I think we will again need to do something similar we did with Nuance? to kind of split the name and code string to fill in these fields on the backend?

```
  {
    name: "Dutsch",
    code: "nl",
    model: "general",
    tier: "enhanced"
  },
```

davehorton: I would like to remove tier and model and simply have
```js
{
    name: "Chinese - general",
    code: "zh"
  },
  {
    name: "Chinese (China)",
    code: "zh-CN"
  },
  {
    name: "Chinese (Taiwan)",
    code: "zh-TW"
  },
```

the choice of "tier" and "model' we will not store as part of the language credential, but rather let applications select them at run time

EgleHelms: clear, will add all the languages.

EgleHelms: I still have one language that has the same code "Flemish" has also an "nl" code for some reason : https://developers.deepgram.com/documentation/features/language/ This throws errors in the console in the front end.

src/vendor/speech-recognizer/deepgram-speech-recognizer-lang.ts
```diff
@@ -0,0 +1,132 @@
...
+    name: "English (New Zealand)",
+    code: "en-NZ",
+  },
+  {
+    name: "English (United States)",
+    code: "en-US",
+  },
+  {
+    name: "Flemish - general",
+    code: "nl",
```
EgleHelms, line 48: @davehorton this code of the langauge creates a problem and console throws an error in the frontend when listing this.
@kitajchuk Shall I use something else for indexing?

src/vendor/speech-recognizer/deepgram-speech-recognizer-lang.ts
davehorton: Weird.  I'll try to open an issue with deepgram, for now just remove "flemish" and leave "Dutch"

src/vendor/speech-recognizer/deepgram-speech-recognizer-lang.ts
davehorton: just eliminate flemish for now from the list

EgleHelms: took it out for now.

EgleHelms: took it out for now.

EgleHelms:
@davehorton @kitajchuk please review this PR.

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -379,8 +386,10 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
             />
           </fieldset>
         )}
-        {/* Single API key state var is used for both Microsoft and WellSaid */}
-        {(vendor === VENDOR_MICROSOFT || vendor === VENDOR_WELLSAID) && (
+        {/* Single API key state var is used for both Microsoft, WellSaid */}
```
kitajchuk, line 389: Ha this comment can be removed now since we’re adding more and more vendors.

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -137,7 +138,9 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
           access_key_id: vendor === VENDOR_AWS ? accessKeyId : null,
           secret_access_key: vendor === VENDOR_AWS ? secretAccessKey : null,
           api_key:
-            vendor === VENDOR_MICROSOFT || vendor === VENDOR_WELLSAID
+            vendor === VENDOR_MICROSOFT ||
```
kitajchuk, line 141: This may be getting out of hand, especially if we add more vendors so may consider extracting some simple utils for these checks at some point. Okay for now tho 🤔

src/vendor/index.tsx
```diff
@@ -83,6 +88,7 @@ export const useSpeechVendors = () => {
       import("./speech-synthesis/ms-speech-synthesis-lang"),
       import("./speech-synthesis/wellsaid-speech-synthesis-lang"),
       import("./speech-synthesis/nuance-speech-synthesis-lang"),
+      import("./speech-recognizer/deepgram-speech-recognizer-lang"),
```
kitajchuk, line 91: Nit but I would group this with the other recognizers

kitajchuk (approved):
Some minor notes but looks good 👍

src/containers/internal/views/speech-services/form.tsx
davehorton: yes, very shortly we will be adding support for ibm watson, as well as a generic api endpoint that would let a third party implement their own connection, so the list will be growing

kitajchuk: Ok cool. Maybe makes sense on the next vendor add to look at some simple utilities for these growing checks to pull clutter out of the component 👍

EgleHelms:
applied the comments from above.

PR #152, opened 2022-11-17
Scope specific user dashboard

EgleHelms:
- added scope selection in the userForm
- limited views for different scopes
- improved userForm with hiding temporary password 
- solved some bugs from parent branch


src/containers/internal/views/users/form.tsx
```diff
@@ -30,12 +42,15 @@ export const UserForm = ({ user }: UserFormProps) => {
   const { signout } = useAuth();
   const navigate = useNavigate();
   const currentUser = useSelectState("user");
+  const currentServiceProvider = useSelectState("currentServiceProvider");
+  const [accounts] = useServiceProviderData<Account[]>("Accounts");
   const [pwdSettings] =
     useApiData<PasswordSettings>("PasswordSettings") || DEFAULT_PSWD_SETTINGS;
```
EgleHelms, line 48: @kitajchuk I think I need your short help here a bit. This is a general question, but I was not sure where to ask it but here. Just FYI this PR is not ready yet and does not have all the code changes I have locally.

So my current issue is:
When I log in as a user that is not admin, I see a ton of errors of the api calls made that cannot be made by other then admin e.g. in this line receiving password settings. All of these calls are usually made in the beginning of the react functional element. Do all of them need to be conditioned according to the scope then or is there a way to make do this smarter?

EgleHelms: I actually think this comes from the backend. So I think password settings should be allowed to fetch for any user existing user that is logged in?

and other stuff like service provider data, should be limited to that one particular SP.

davehorton: agree, anyone can retrieve password settings

kitajchuk: Do you have a list of all the api calls that don’t work for non admins? My guess is if we’re now getting into user scope and permissions it’s time to get that stuff properly implemented. We have a basic ACL component but we’re going to need to make some modifications to how that works because it’s currently sort of dumb and was designed to work for admins and whether or not to show the MS Teams link. I’ve articulated a lot of thinking around proper permissions implementation to @davehorton so I can surface that and share the thinking. 

I have a big BUT here, however. I don’t think it’s wise to confuse the higher level permissions work with this user dashboard stuff. Especially since these PRs are big enough already. My take is we should wrap the first PR that is scoped to admin only and then we should create and document an issue for the proper permissions implementation, which when done, would ideally include making sure user permissions are accounted for here on this page.

I wouldn’t want to see us approach this any other way really because otherwise it’s too messy being mixed up in this stuff here. 

@davehorton @EgleHelms thoughts?

EgleHelms: @kitajchuk thanks for your feedback. Would you like us to merge the other PR, or do some other changes there? I believe your comments were addressed there. So I moved on to implement the scoped user creation and only when logging in as scoped users that I noticed the permission stuff mentioned above coming up. 

Would you like me to include the scoped user creation into the 1st PR and keep this PR or start a new one for the permission things? 

Cause to my understanding we cannot merge the 1st PR into the main until the whole multiuser feature + permissions are implemented, or? ...so I branched off and continued since as you said, the first PR is already pretty big. Sorry, if this was too soon. Let me know your instructions.

davehorton: Yes I would like to merge that other PR with just the admin-level user work, then move on.

PR #154, opened 2022-11-24
ScopeAccess component implementation

EgleHelms:
@kitajchuk this is my try to implement what you described with a simple component, plus applied it as an example in one place which works well in the app itself. I also added a similar cypress test, but that one does not render for some reason (I find that strange since it works in the app), maybe you can spot why with your pro eyes :).

I was not exactly sure about the hook and what exactly you meant about that. Maybe you can let me know where we would use it. Now I didn't think this was nescesarry.

Let me know your thoughts. And many thanks for guiding me.

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -1,5 +1,4 @@
 import type { Vendor } from "src/vendor/types";
-import type { Scope } from "src/store/types";
 
 /** Simple types */
 
@@ -117,7 +116,6 @@ export interface User {
   service_provider_sid?: string | null;
   initial_password?: string;
   permissions?: UserPermissions;
-  enumScope: Scope;
 }
 
 export interface UserLogin extends User {
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -1,4 +1,5 @@
 import type { Vendor } from "src/vendor/types";
+import type { Scope } from "src/store/types";
 
 /** Simple types */
 
@@ -116,6 +117,7 @@ export interface User {
   service_provider_sid?: string | null;
   initial_password?: string;
   permissions?: UserPermissions;
+  enumScope: Scope;
 }
 
 export interface UserLogin extends User {
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -19,13 +19,6 @@ export type UserPermissions =
   | "PROVISION_SERVICES"
   | "PROVISION_USERS";
 
-// We'll want something like this for actual permissions implementation...
-// export enum UserPermissions {
-//   VIEW_ONLY = 0,
-//   PROVISION_SERVICES = 1,
-//   PROVISION_USERS = 2,
-// }
-
 /** Status codes */
 
 export enum StatusCodes {
@@ -106,21 +99,24 @@ export interface PasswordSettings {
 /** API responses/payloads */
 
 export interface User {
-  scope?: UserScopes;
+  scope: UserScopes;
   user_sid: string;
   name: string;
   email: string;
   is_active: boolean;
   force_change: boolean;
-  account_sid?: string | null;
-  service_provider_sid?: string | null;
+  account_sid: string | null;
+  service_provider_sid: string | null;
   initial_password?: string;
-  permissions?: UserPermissions;
+  permissions?: UserPermissions[];
 }
 
-export interface UserLogin extends User {
+export interface UserLogin {
   token: string;
   force_change: boolean;
+  scope: UserScopes;
+  user_sid: string;
+  permissions: UserPermissions[];
 }
 
 export interface UserLoginPayload {
@@ -138,6 +134,14 @@ export interface UserUpdatePayload {
   is_active: boolean;
   service_provider_sid: string | null;
   account_sid: string | null;
+}
+
+export interface UserJWT {
+  scope: UserScopes;
+  user_sid: string;
+  account_sid?: string | null;
+  service_provider_sid?: string | null;
+  permissions: UserPermissions[];
 }
 
 export interface ServiceProvider {
--- a/src/components/scoped-access.cy.tsx
+++ b/src/components/scoped-access.cy.tsx
@@ -1,0 +1,43 @@
+import React from "react";
+import { H1, H2 } from "jambonz-ui";
+
+import { TestProvider } from "src/test";
+import { ScopedAccess } from "./scoped-access";
+
+import type { ScopedAccessProps } from "./scoped-access";
+
+/** Wrapper to pass different user scopes as enum values */
+const ScopedAccessTestWrapper = (props: Partial<ScopedAccessProps>) => {
+  return (
+    <TestProvider>
+      <ScopedAccess
+        enumScope={props.enumScope!}
+        noAccessRender={<H2>{"noAccessRender"}</H2>}
+      >
+        <div className="scope-div">
+          <H1>ScopedAccess: {props.enumScope}</H1>
+        </div>
+      </ScopedAccess>
+    </TestProvider>
+  );
+};
+
+describe("<ScopedAccess>", () => {
+  it("has sufficient scope - admin", () => {
+    cy.mount(<ScopedAccessTestWrapper enumScope={0} />);
+    cy.get(".scope-div").should("exist");
+    cy.get("h1").should("exist");
+  });
+
+  it("has insufficient scope - service_provider", () => {
+    cy.mount(<ScopedAccessTestWrapper enumScope={1} />);
+    cy.get(".scope-div").should("not.exist");
+    cy.get("h2").should("exist");
+  });
+
+  it("has insufficient scope - account", () => {
+    cy.mount(<ScopedAccessTestWrapper enumScope={2} />);
+    cy.get(".scope-div").should("not.exist");
+    cy.get("h2").should("exist");
+  });
+});
--- a/src/components/scoped-access.cy.tsx
+++ b/src/components/scoped-access.cy.tsx
@@ -1,21 +1,19 @@
 import React from "react";
-import { H1, H2 } from "jambonz-ui";
+import { H1 } from "jambonz-ui";
 
 import { TestProvider } from "src/test";
 import { ScopedAccess } from "./scoped-access";
 
 import type { ScopedAccessProps } from "./scoped-access";
+import { Scope } from "src/store/types";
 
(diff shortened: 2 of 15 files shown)


src/store/types.ts
```diff
@@ -14,6 +14,12 @@ export interface ACL {
   hasMSTeamsFqdn: boolean;
 }
 
+export enum Scope {
```
kitajchuk, line 17: This should be reversed so that admin is the highest numeric value

src/store/actions.ts
```diff
@@ -3,7 +3,8 @@ import { sortLocaleName } from "src/utils";
 import { getToken, parseJwt } from "src/router/auth";
 
 import type { State, Action } from "./types";
-import type { ServiceProvider, User } from "src/api/types";
+import { Scope } from "./types";
```
kitajchuk, line 6: Idk why the imports changed like this. You could have just added Scope to the type imports above and there was no need to change the other one.

src/containers/internal/navi/index.tsx
```diff
@@ -158,14 +159,16 @@ export const Navi = ({
...
+          <ScopedAccess enumScope={0}>
+            <button
+              type="button"
+              onClick={() => setModal(true)}
+              title="Add service provider"
+              className="btnty"
+            >
+              <Icons.PlusCircle />
+            </button>
+          </ScopedAccess>
```
kitajchuk, line 171: Can you do me a favor and remove the below stuff regarding the user that is intentionally not rendered. This is old and won’t be used ever. You can also remove react-blockies from the package.json 👍

src/containers/internal/navi/index.tsx
```diff
@@ -149,7 +150,7 @@ export const Navi = ({
                     );
                   })
                 ) : (
-                  <option>&nbsp;</option>
+                  <option>currentServiceProvider.name</option>
```
kitajchuk, line 153: Why? This was intentional since the current SP is reliant on an async request. In fact, the ternary implies that it’s not set, so undefined, in this else block so this should not have been changed…

src/components/scoped-access.tsx
```diff
@@ -0,0 +1,29 @@
...
+  enumScope,
+  children,
+  noAccessRender,
+}: ScopedAccessProps) => {
+  const user = useSelectState("user");
+
+  if (user && user.enumScope <= enumScope) {
+    return <>{children}</>;
+  }
+
```
kitajchuk, line 23: I don’t think we need this so it can be removed. You either have access to things or you don’t and there’s no reason to show people a message about what they can’t access.

src/components/scoped-access.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+
+import { useSelectState } from "src/store";
+
+import type { Scope } from "src/store/types";
+
+export type ScopedAccessProps = {
+  enumScope: Scope;
```
kitajchuk, line 8: Would prefer to just call this prop scope

src/components/scoped-access.cy.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+describe("<ScopedAccess>", () => {
+  it("has sufficient scope - admin", () => {
+    cy.mount(<ScopedAccessTestWrapper enumScope={0} />);
+    cy.get(".scope-div").should("exist");
+    cy.get("h1").should("exist");
+  });
+
+  it("has insufficient scope - service_provider", () => {
+    cy.mount(<ScopedAccessTestWrapper enumScope={1} />);
```
kitajchuk, line 33: Anywhere you use the scope prop you should be using the enum rather than raw numbers as a value

src/api/types.ts
```diff
@@ -116,6 +117,7 @@ export interface User {
   service_provider_sid?: string | null;
   initial_password?: string;
   permissions?: UserPermissions;
+  enumScope: Scope;
```
kitajchuk, line 120: I don’t think we should put this here since this is supposed to be an interface for the api response and this enum is purely a client-side thing.

src/store/actions.ts
```diff
@@ -63,7 +64,8 @@ export const currentServiceProviderAction = (
 
 export const userAsyncAction = async (): Promise<User> => {
   const token = getToken();
-  return parseJwt(token);
+  const userData = parseJwt(token);
+  return { ...userData, enumScope: Scope[userData.scope] };
```
kitajchuk, line 68: Maybe we call this scopeAccess or something, I don’t like calling it enum since that would be like saying “varScope” you know…

src/api/types.ts
kitajchuk: For the store just maybe have a unique user interface since we only have the JWT data there anyway.

kitajchuk (changes requested):
Some changes. For the tests to work you will have to mock the user scope and login since that’s not the default. You can see the test utils here: https://github.com/jambonz/jambonz-webapp/blob/main/src/test/index.tsx. This may be a bit more challenging since we don’t currently do a login example in any of our component tests.

We don’t currently have any flows for this. To test scopes we’d need to be able to mock the store more flexibly for our tests.

You notice currently we allow passing auth context props but we use the store provider verbatim. We’d need to use the store context similar to how we do with auth in the test provider and support passing store props as well. May be more trouble than it’s worth for this PR but you could take a shot st it if you want. I would separate them out as authProps and storeProps on the test provider. If you do you may have to update other tests to pass props the new way.

src/store/actions.ts
```diff
@@ -63,7 +64,8 @@ export const currentServiceProviderAction = (
 
 export const userAsyncAction = async (): Promise<User> => {
   const token = getToken();
-  return parseJwt(token);
+  const userData = parseJwt(token);
```
kitajchuk, line 67: As I was mentioning above you would change the user type in the store as described and use it here as well as the type passed to the Promise generic

src/api/types.ts
kitajchuk: To me it also looks like you need to add all this user login properties to the cypress fixture for it as well please.

src/containers/internal/navi/index.tsx
EgleHelms: yeah, that was not smart. hmm, then I guess I need another condition here, to map the values when it is only the currentServiceProvider is available e.g. for SP and account users, they wont see other ones.

src/store/actions.ts
EgleHelms: hmm, here I separated them because using it like this: scopeAccess: Scope[userData.scope] is not allowed if it is used as a type, since then it does not pick up the value, just type. Do you know a better way to do this?

src/components/scoped-access.tsx
EgleHelms: Ahm, it was not about showing some message, but more like if you pass a noAccessRender component, it can render something else for different scopes. 
E.g. usage I was thinking this for was - when clicking on "Accounts" in the Nav admin  and SP scope user will see Accounts list, but I would render "edit page" the account scope user for the account he is part of. And then if there is no noAccessRender component passed, then I would then just use null
Does it make sense?

kitajchuk: I think pieces of UI should be wrapped vs this approach if I’m understanding correctly. Just wrap the edit button with the scoped access.

src/store/actions.ts
kitajchuk: Ha ya that makes sense. I always forget that with enums. My preference for specifying type imports is probably annoying for this reason 🙃

src/api/types.ts
EgleHelms: if I update the cypress fixture, I cannot resolve the typing of UserLogin in the test/index.tsx I am not sure why. because that is an extension of User interface. it has all the relevant fields still complains somehow.

EgleHelms:
> Some changes. For the tests to work you will have to mock the user scope and login since that’s not the default. You can see the test utils here: https://github.com/jambonz/jambonz-webapp/blob/main/src/test/index.tsx. This may be a bit more challenging since we don’t currently do a login example in any of our component tests.
> 
> We don’t currently have any flows for this. To test scopes we’d need to be able to mock the store more flexibly for our tests.
> 
> You notice currently we allow passing auth context props but we use the store provider verbatim. We’d need to use the store context similar to how we do with auth in the test provider and support passing store props as well. May be more trouble than it’s worth for this PR but you could take a shot st it if you want. I would separate them out as authProps and storeProps on the test provider. If you do you may have to update other tests to pass props the new way.

I was trying to resolve this but wasn't able to wrap my head around how to pass the values correctly. I will try again Monday. Feel free to push into this branch anything you want to help move this forward. Thanks.

src/test/index.tsx
```diff
@@ -33,8 +46,15 @@ export const authProps: AuthStateContext = {
...
+  const userJWT = parseJwt(userLogin.token);
+  const [state, dispatch]: [State, React.Dispatch<Action<keyof State>>] =
+    useReducer(reducer, { ...initialState, user: userJWT });
+
+  const globalDispatch: GlobalDispatch = middleware(dispatch);
+  const storeProps: AppStateContext = { state, dispatch: globalDispatch };
+
   return (
-    <StateProvider>
+    <StateContext.Provider value={storeProps}>
```
EgleHelms, line 57: @kitajchuk I try to pass the store here, and it should be part of the "...restProps" but they don't travel into the Wrapper of the test. I don't understand why.

kitajchuk: This isn’t really how I was imagining this implementation here. It can be much simpler (we don’t have to replicate the useReducer stuff—that’s overkill. And the userJWT should not be here. You would set your user in the tests. Here we just want to “mock” it in so we can pass ad-hoc props to set global state and test components. I also wouldn’t merge auth props and store props like this because it is messy. I was thinking separating them on the props type as authProps and storeProps and spreading them separately on each context. Maybe just revert this commit and I can branch off this and show you what I was thinking 🤔

EgleHelms: @kitajchuk reverted.

EgleHelms:
@kitajchuk @davehorton  can this be merged with your changes added?

kitajchuk (approved):

kitajchuk:
> > @kitajchuk @davehorton can this be merged with your changes added?
> 
> What changes were those? I dont think I have any changes on the main branch of this repo

I believe yes since you merged my PR into this branch 👍

PR #159, opened 2022-11-30
feature/scope limits

EgleHelms:
Limitations considered and should function:

**_When a service provider user logs into the system, the UI must be modified for them in the following aspects:_**

- +The service provider dropdown at the top of the view is removed.
- +The Settings panel includes only the Service Provider tab
- +The User dashboard allows the logged-in user to view only users associated with the current service provider
- +to create other service provider user accounts (for the same service provider) 
- +to delete accounts for the same service provider.
- +the data shown in all other panels shall be restricted to the current service provider. 
- +service provider users shall create/edit/delete new users at the account level, by selecting an existing account that falls under their service provider.

**_When an account-level user logs into the system, the UI shall be modified for them in the following aspects:_**

- +The Settings panel is removed
- +The account list is removed (since there is only one account)
- +A new Account panel is available on the left-hand side menu, it shows the current Add/Edit the current user Account panel
- +account users shall create/edit/delete new users at the account level, that falls under their service provider +account.
- +Recent Calls and Alerts are filtered to show only the calls and alerts for this account
- +The carrier panel continues to list all of the carriers defined for the account or service provider
- +Carriers can only be added at the account level.
- +only the account-level carriers can be edited or deleted.  
- +The speech panel continues to list all of the speech credentials
- +only the account level credentials can be edited or deleted.


src/containers/internal/views/speech-services/form.tsx
```diff
@@ -258,10 +275,16 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
...
+                ? accounts?.filter(
+                    (acct) => acct.account_sid === user.account_sid
+                  )
+                : accounts
+            }
             account={[accountSid, setAccountSid]}
             required={false}
-            defaultOption
+            defaultOption={user?.scope !== USER_ACCOUNT}
             disabled={credential ? true : false}
```
EgleHelms, line 288: @kitajchuk @davehorton why are we not allowed to make changes to already created credentials?

davehorton: we only allow you to update something like region.  Other than that you should destroy and create a new credential.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -437,6 +440,11 @@ export const CarrierForm = ({
   const handleSubmit = (e: React.FormEvent) => {
     e.preventDefault();
 
+    if (user?.scope === USER_ACCOUNT && user.account_sid !== accountSid) {
```
EgleHelms, line 443: @kitajchuk  I am really having trouble thinking of a good check here + when the defaultOption could be used for choosing Account in AccountSelect. If Account Select was correct, this would also be correct

src/containers/internal/views/carriers/form.tsx
```diff
@@ -677,11 +696,17 @@ export const CarrierForm = ({
...
+                    ? accounts?.filter(
+                        (acct) => user.account_sid === acct.account_sid
+                      )
+                    : accounts
+                }
                 account={[accountSid, setAccountSid]}
                 label="Used By"
                 required={false}
-                defaultOption
+                defaultOption={user?.scope !== USER_ACCOUNT}
```
EgleHelms, line 709: @kitajchuk I mean this. The premise is:
- user. scope is account and page is edit carrier is for one account, there should be no default option
- user.scope is account and page is add, there should be no default option
- user.scope is account and page is edit and carrier is meant for all accounts, there should be default option "All Accounts"

src/containers/internal/navi/index.tsx
```diff
@@ -172,9 +174,47 @@ export const Navi = ({
...
           <ul>
-            {naviTop.map((item) => (
-              <Item key={item.label} item={item} handleMenu={handleMenu} />
-            ))}
+            {naviTop.map((item) => {
+              if (item.label === "Settings") {
+                return (
+                  <ScopedAccess scope={Scope.service_provider} user={user}>
+                    <Item
+                      key={item.label}
```
EgleHelms, line 182: @kitajchuk getting a key issue in the console, but to me, it seems like I do take care that the keys are unique.

src/containers/login/create-password.tsx
```diff
@@ -65,7 +72,11 @@ export const CreatePassword = () => {
           if (response.status === StatusCodes.NO_CONTENT) {
             sessionStorage.clear();
 
-            navigate(ROUTE_INTERNAL_ACCOUNTS);
+            navigate(
+              user?.scope !== USER_ACCOUNT
```
EgleHelms, line 76: @kitajchuk this routing doesn't work here, but does in login.tsx. I get the user state already logged and the check is corrrect. Maybe you can advise me.

EgleHelms:
@kitajchuk there is also some error in the console about icons being defined popping up sometimes from chunk.js. I am not too sure what is causing it. Seems like an unresolved promise.

EgleHelms:
@kitajchuk I am really stuck with the things above. Not sure if I am working too long on this, but my brain is fried, seems like simple things, but I cannot find solutions. Let me know if you have advice on any of the above. Thanks a lot.

complementary branch on api server: https://github.com/jambonz/jambonz-api-server/pull/80

src/containers/internal/navi/index.tsx
EgleHelms: This was caused by the ScopedAccess component it also needed a key.

src/containers/internal/views/carriers/form.tsx
EgleHelms: I added a function to add all these checks, could not find a better way then this. Maybe you can advise.

davehorton:
It seems like when I am logged in as 'admin' user I am unable to create a Carrier at the account level ("Used by" is set to all accounts and is disabled.

image

davehorton:
As an account level user (for the "default account") I added a carrier (twilio) and it said it was saved succesfully.  However, it then does not appear in the carrier list for that user

image

davehorton:
As an account level user, when I log in and choose the Speech Services list, the dropdown includes "all accounts", "default account" -- this is the associated account of the logged in user, but also includes "account 2" which is a totally different account that I created under this SP.  This is wrong, an account level user should not be able to any other account appear in any list view or dropdown -- ie it should be considered "off limits" for an account user to know of the existence of any other accounts under his/her service provider

Note that the similar dropdown for the Carriers list *does* seem to be filtered properly (ie for my user showed only "all accounts" and "default account")

src/containers/login/create-password.tsx
EgleHelms: I had to use parseJwt here. The user state is not correct after logging out as admin.

EgleHelms:
@davehorton I addressed all the issues above, please re-review.

src/containers/login/create-password.tsx
kitajchuk: The user hasn’t been dispatched yet at this point.

davehorton:
as 'admin' user, I created another admin user (admin2).  The user was saved, but the "edit user" window immediately opened for that new admin user, however scope showed "Service Provider"
image

davehorton:
The text of the labels for the checkboxes below for custom speech for Microsoft azure should be changed to
- "Use a custom voice"
- "Use a custom speech model"

image

davehorton:
As an account level user I should not be able to view sip or smpp password information for a carrier created at the service provider level.

image

and

image


I should not have been able to view either of those.  The easiest way to hide might be to disable or hide the button on the right side of the password fields in both forms

davehorton:
Now that we have multiple users support, I think we should show somewhere the username of the currently logged in user along with their scope.

davehorton:
I think on the users list view we should add a column to the right, which is populated as follows
- account scope: show the account name
- service provider scope: show the service provider name
- admin scope: leave blank

Currently, the view looks like this below, which will make it not so useful when we have lots of users across different SPs and accounts

Additionally, it should be easy to filter the view for a single service provider or account by name

image

EgleHelms:
> * Use a custom

I agree with showing the account name for account users it belongs to and just by listing it the fuzzy search will pick it up, so that is easy enough to add.
But service provider is the always the current.

davehorton:
> "There are also authentication fields in Account/Application, any of those also need disabling?"

Thats a really good question.  Applications are at the account level, so if you are an account user looking at an application, it must be "your" application so I think we are all set there.  Ditto accounts -- I think as an account user you are all set.

So I suppose the question is, should a service provider be able to see the applications that an account has set up, and see the api keys.  And possibly the http basic auth that an account has set up.

On the one hand, there is an argument to be made that a service provider should not be able to see, for instance, the api keys that an account has created, because otherwise they could use them maliciously somehow.

On the other hand, it will be a common case that service providers will want to handle all of the provisioning for sub-accounts.

Given those two competing scenarios, I am ok leaving it as currently working -- service providers can have all rights (e.g. create and view api keys) for their sub-accounts.

src/utils/use-scoped-redirect.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from "react";
+import { useNavigate } from "react-router-dom";
+
+import { toastError } from "src/store";
+
+import type { IMessage, Scope, UserData } from "src/store/types";
+
+export const useScopedRedirect = (
```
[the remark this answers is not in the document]
kitajchuk: Okay. Sorry I’ve been busy and unable to review. I will make time today (I meant to yesterday but got caught up).

EgleHelms:
> I see this API failing in console—what's up with that? I am using the user api branch for the server backend...
> 
> Screen Shot 2022-12-08 at 8 08 29 AM

@kitajchuk waiting for @davehorton to push this fix, since he did not want to jwt encode the username, he will fix the Users/me to be used by all systems not only hosted ones.

davehorton:
> > I see this API failing in console—what's up with that? I am using the user api branch for the server backend...
> > Screen Shot 2022-12-08 at 8 08 29 AM
> 
> @kitajchuk waiting for @davehorton to push this fix, since he did not want to jwt encode the username, he will fix the Users/me to be used by all systems not only hosted ones.

Yes, that is because that particular API has only been used till now in the hosted version of jambonz, where every user is an account user.  I am going to fix that shortly (today)

davehorton:
please pull the latest jambonz-api-server on the feature/user-api-calls to resolve the issues calling /Users/me @kitajchuk  @EgleHelms

kitajchuk:
@EgleHelms can you make sure your problematic redirect code is pushed either here or on another branch pointing at this one so I can troubleshoot your problem? Thanks 🙏

kitajchuk:
@EgleHelms @davehorton I put up a PR against this branch with the changes to get that scoped redirect hook working.

https://github.com/jambonz/jambonz-webapp/pull/164

Really the only issue there was that since the `user` coming from our store state can be undefined, e.g. `user?: UserData;` we need to account for this in the hook implementation. Optional parameters cannot come before required parameters when typing function arguments so I simply moved the args order. Then with the hooks effect we just need to actually check that user  is not `undefined` before proceeding with our logic and possible redirect 👍

davehorton:
hi @EgleHelms  @kitajchuk what is the status of this PR / branch from your perspectives?  I see it needs to be rebased on main, but is it otherwise ready to merge? What is outstanding?

EgleHelms:
> hi @EgleHelms @kitajchuk what is the status of this PR / branch from your perspectives? I see it needs to be rebased on main, but is it otherwise ready to merge? What is outstanding?

Still things I am finishing-up/working on:
1) adding redirect restriction according to resource not only route (e.g. manually entering links with account sid will not work if the user is not the owner of that resource)
2) Idea of merging Carrier/Speech lists to include both all and account-specific entries from Account users - @davehorton you can let me know if I should address this in this branch
3) Still need to fix the Links for Carriers that were doubling (have a fix but did not commit because of the point above that is related)

davehorton:
I do want to do your point 2 above, but lets merge this branch after 1/3 are done and then create a new feature branch for that.

EgleHelms:
I don't know how to rebase this without merging  main.

kitajchuk:
> I am sorry, I don't know how to rebase this without merging main into this branch, needed to undo one faulty try to rebase the conflict that occurred.

The button on GitHub let’s you choose to rebase update from main. Otherwise I always do it locally: `git pull origin main --rebase`

EgleHelms:
> > I am sorry, I don't know how to rebase this without merging main into this branch, needed to undo one faulty try to rebase the conflict that occurred.
> 
> The button on GitHub let’s you choose to rebase update from main. Otherwise I always do it locally: `git pull origin main --rebase` from your branch…

Yeah, the commands I know, but I initialize a rebase and the button said there is a conflict, fix the conflict and try to complete the rabase and I then get a message that I should still pull changes, and that results in another merge of main. When I do pull origin main before the rebase it says my branch is up to date. If I say don't pull after the rebase I get a git error. So I dont know how to fix that without a merge.

kitajchuk:
> > > I am sorry, I don't know how to rebase this without merging main into this branch, needed to undo one faulty try to rebase the conflict that occurred.
> > 
> > 
> > The button on GitHub let’s you choose to rebase update from main. Otherwise I always do it locally: `git pull origin main --rebase` from your branch…
> 
> Yeah, the commands I know, but I initialize a rebase and the button said there is a conflict, fix the conflict and try to complete the rabase and I then get a message that I should still pull changes, and that results in another merge of main. When I do pull origin main before the rebase it says my branch is up to date. If I say don't pull after the rebase I get a git error. So I dont know how to fix that without a merge.

When you rebase locally (e.g. resolve conflicts) you have you force push your branch back up because you’ve changed the history of your branch otherwise you will have this issue you are describing. Merges of main or your own branch into your branch are not desirable. If your local branch feels borked you can always get your state prior to the rebase/merge issues from git reflog.

EgleHelms:
@kitajchuk I think I was successful in rebasing. Thanks for the help. So the procedure should be rebasing every day starting work to always be up to date with the main (if it is such a long PR)? Is this the result that I am not doing it enough? 

I think now this is ready for the last reviews.

kitajchuk:
> @kitajchuk I think I was successful in rebasing. Thanks for the help. So the procedure should be rebasing every day starting work to always be up to date with the main (if it is such a long PR)? Is this the result that I am not doing it enough?
> 
> I think now this is ready for the last reviews.

Not necessarily. You can rebase whenever you feel like it. In this case I’m guessing you got conflicts with files from the other scoped limits work that was merged. It happens. No big deal.

davehorton (approved):

davehorton:
I've tested and it looks good to me.  @kitajchuk can you give a review and let us know if if this is good to go?

kitajchuk:
> I've tested and it looks good to me. @kitajchuk can you give a review and let us know if if this is good to go?

I’ll check it out now 🤔

src/containers/internal/views/carriers/index.tsx
```diff
@@ -39,15 +49,30 @@ export const Carriers = () => {
...
       : [];
-  }, [accountSid, carriers]);
+  }, [accountSid, carrier, carriers]);
 
   const filteredCarriers = useFilteredResults<Carrier>(
     filter,
     carriersFiltered
   );
 
+  const getCarriers = (url: string) => {
```
kitajchuk, line 59: I’ve noticed we’ve done this a number of times where we scrap the hook when we need a conditional api url. Not ideal IMO. I think it would be better to add the url as a dependency in the hooks effect so that when it changes a refetch will happen. I think this could be leveraged on the useApiData hook but doesn’t make sense in the useCurrentServiceProviderData hook 🤔

src/containers/internal/views/carriers/index.tsx
```diff
@@ -91,6 +124,20 @@ export const Carriers = () => {
     }
   };
 
+  useEffect(() => {
+    if (accountSid) {
```
kitajchuk, line 128: See it would be less code to update a url in local state and let the hook do all the work. You could reduce a lot of this code here.

src/containers/internal/views/carriers/index.tsx
```diff
@@ -39,15 +49,30 @@ export const Carriers = () => {
...
+        setCarriers(json);
+      })
+      .catch((error) => {
+        toastError(error.msg);
+      });
+  };
+
   const handleDelete = () => {
     if (carrier) {
+      if (user?.scope === USER_ACCOUNT && user.account_sid !== accountSid) {
```
kitajchuk, line 71: Same

src/containers/internal/views/carriers/index.tsx
```diff
@@ -109,7 +156,13 @@ export const Carriers = () => {
         />
         <AccountFilter
           account={[accountSid, setAccountSid]}
-          accounts={accounts}
+          accounts={
```
kitajchuk, line 159: Same

src/containers/internal/views/phone-numbers/form.tsx
```diff
@@ -88,6 +88,7 @@ export const PhoneNumberForm = ({ phoneNumber }: PhoneNumberFormProps) => {
         .then(() => {
           phoneNumber.refetch();
           toastSuccess("Phone number updated successfully");
+          navigate(ROUTE_INTERNAL_PHONE_NUMBERS);
```
kitajchuk, line 91: Idk why this is desirable but someone prefers it?

src/containers/internal/views/phone-numbers/index.tsx
```diff
@@ -123,8 +125,14 @@ export const PhoneNumbers = () => {
         />
         <AccountFilter
           account={[accountSid, setAccountSid]}
-          accounts={accounts}
-          defaultOption
+          accounts={
```
kitajchuk, line 128: Same

src/containers/internal/views/recent-calls/index.tsx
```diff
@@ -86,7 +91,13 @@ export const RecentCalls = () => {
       <section className="filters filters--multi">
         <AccountFilter
           account={[accountSid, setAccountSid]}
-          accounts={accounts}
+          accounts={
```
kitajchuk, line 94: Same

src/containers/internal/views/speech-services/edit.tsx
```diff
@@ -1,22 +1,43 @@
...
+import { useNavigate, useParams } from "react-router-dom";
 
 export const EditSpeechService = () => {
   const params = useParams();
-  const [data, refetch, error] = useServiceProviderData<SpeechCredential>(
-    `SpeechCredentials/${params.speech_credential_sid}`
+  const navigate = useNavigate();
+  const user = useSelectState("user");
+
+  const [data, refetch, error] =
```
kitajchuk, line 20: You cannot do this. This is breaking the rules of hooks for React. You have to call hooks in the same order in your components so this is just plain disallowed. I feel like you should be getting a warning about this? This reassures my desire to update the useApiData hook so we can update the url we pass in and change our data. So this 100% needs to be addressed

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -98,6 +100,13 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
   const handleSubmit = (e: React.FormEvent) => {
     e.preventDefault();
 
+    if (user?.scope === USER_ACCOUNT && user.account_sid !== accountSid) {
```
kitajchuk, line 103: Same

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -136,40 +146,55 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
...
-          secret: vendor === VENDOR_NUANCE ? secretKey : null,
-          stt_api_key: sttApiKey || null,
-          stt_region: sttRegion || null,
-          tts_api_key: ttsApiKey || null,
-          tts_region: ttsRegion || null,
-        })
-          .then(({ json }) => {
-            toastSuccess("Speech credential created successfully");
-            navigate(`${ROUTE_INTERNAL_SPEECH}/${json.sid}/edit`);
+        if (user)
```
kitajchuk, line 156: Unnecessary with api implementation changes

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -136,40 +146,55 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
...
+              navigate(`${ROUTE_INTERNAL_SPEECH}/${json.sid}/edit`);
+            })
+            .catch((error) => {
+              toastError(error.msg);
+            });
+      }
+    }
+  };
+
+  const checkOptions = () => {
```
kitajchuk, line 185: You can make the same changes here. Also you could make this a util that takes an object since we look at the same data properties as with the same concept on the carrier view

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -289,10 +314,16 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
         </fieldset>
         <fieldset>
           <AccountSelect
-            accounts={accounts}
+            accounts={
```
kitajchuk, line 317: Same

src/containers/internal/views/speech-services/index.tsx
```diff
@@ -1,25 +1,53 @@
...
   const currentServiceProvider = useSelectState("currentServiceProvider");
   const [accounts] = useServiceProviderData<Account[]>("Accounts");
   const [accountSid, setAccountSid] = useState("");
   const [credential, setCredential] = useState<SpeechCredential | null>(null);
   const [credentials, setCredentials] = useState<SpeechCredential[]>();
+  const [filter] = useState("");
+
+  const credentialsFiltered = useMemo(() => {
+    return credentials
+      ? credentials.filter((credentials) =>
```
kitajchuk, line 39: This arg should be credential singular

src/containers/internal/views/speech-services/index.tsx
```diff
@@ -33,14 +61,22 @@ export const SpeechServices = () => {
 
   const handleDelete = () => {
     if (credential && currentServiceProvider) {
+      if (user?.scope === USER_ACCOUNT && user.account_sid !== accountSid) {
```
kitajchuk, line 64: Same

src/containers/internal/views/speech-services/index.tsx
```diff
@@ -62,13 +98,17 @@ export const SpeechServices = () => {
 
   useEffect(() => {
     if (accountSid) {
-      getSpeechCredentials(`${API_ACCOUNTS}/${accountSid}/SpeechCredentials`);
-    } else if (currentServiceProvider) {
       getSpeechCredentials(
-        `${API_SERVICE_PROVIDERS}/${currentServiceProvider.service_provider_sid}/SpeechCredentials`
+        `${API_ACCOUNTS}/${user?.account_sid || accountSid}/SpeechCredentials`
```
kitajchuk, line 102: This is another case where the changes to useApiData hook could again simplify things for us

src/containers/internal/views/speech-services/index.tsx
```diff
@@ -83,29 +123,45 @@ export const SpeechServices = () => {
       <section className="filters filters--ender">
         <AccountFilter
           account={[accountSid, setAccountSid]}
-          accounts={accounts}
+          accounts={
```
kitajchuk, line 126: Same

src/containers/internal/views/speech-services/status.tsx
```diff
@@ -17,10 +18,15 @@ export const CredentialStatus = ({
   cred,
   showSummary = false,
 }: CredentialStatusProps) => {
+  const user = useSelectState("user");
   const [testResult, testRefetch, testError] =
-    useServiceProviderData<CredentialTestResult>(
-      `SpeechCredentials/${cred.speech_credential_sid}/test`
-    );
+    user && user.scope !== USER_ACCOUNT
```
kitajchuk, line 23: Again no—conditional hooks are not allowed 🚫

src/containers/internal/views/users/form.tsx
```diff
@@ -168,6 +170,7 @@ export const UserForm = ({ user }: UserFormProps) => {
       setIsActive(!!user.data.is_active);
       setEmail(user.data.email);
       setScope(getUserScope(user.data));
+      user.data.account_sid ? setAccountSid(user.data.account_sid) : false;
```
kitajchuk, line 173: This line makes no sense. Did you mean to put the ternary inside the function call?

src/containers/internal/views/users/index.tsx
```diff
@@ -28,6 +38,15 @@ export const Users = () => {
   const [accounts] = useServiceProviderData<Account[]>("Accounts");
 
   const usersFiltered = useMemo(() => {
+    //find and add account/sp names to user to improve fuzzy search
```
kitajchuk, line 41: This is not ideal because you’re mutating what should be immutable data. If this is desirable the backend should return the data in this way…

src/containers/internal/views/users/index.tsx
```diff
@@ -55,7 +74,9 @@ export const Users = () => {
     return [];
   }, [accountSid, scopeFilter, users, accounts, currentServiceProvider]);
 
-  const filteredUsers = useFilteredResults<User>(filter, usersFiltered);
+  const filteredUsers = useFilteredResults<User>(filter, usersFiltered)?.sort(
+    (a, b) => sortAlphabetically(a, b)
```
kitajchuk, line 78: You should just pass the util method as the param to sort…

src/containers/internal/views/users/index.tsx
```diff
@@ -82,16 +103,21 @@ export const Users = () => {
         />
         <AccountFilter
           account={[accountSid, setAccountSid]}
-          accounts={accounts}
-          defaultOption={true}
+          accounts={
+            user?.scope === USER_ACCOUNT
```
kitajchuk, line 107: Same

src/containers/login/create-password.tsx
```diff
@@ -65,7 +72,11 @@ export const CreatePassword = () => {
           if (response.status === StatusCodes.NO_CONTENT) {
             sessionStorage.clear();
 
-            navigate(ROUTE_INTERNAL_ACCOUNTS);
+            navigate(
+              parseJwt(token).scope !== USER_ACCOUNT
```
kitajchuk, line 76: Probably would be preferred to parse above rather than on the fly like this

src/router/auth.tsx
```diff
@@ -101,7 +103,11 @@ export const useProvideAuth = (): AuthStateContext => {
               sessionStorage.setItem(SESS_OLD_PASSWORD, password);
               navigate(ROUTE_CREATE_PASSWORD);
             } else {
-              navigate(ROUTE_INTERNAL_ACCOUNTS);
+              navigate(
+                parseJwt(token).scope !== USER_ACCOUNT
```
kitajchuk, line 107: Same would prefer parse at the top of this function block rather than on the fly

src/styles/_lists.scss
```diff
@@ -34,6 +34,7 @@
     display: flex;
     align-items: center;
     grid-gap: ui-vars.$px02;
+    color: ui-vars.$jambonz;
```
kitajchuk, line 37: Why?

src/utils/index.ts
```diff
@@ -148,11 +149,26 @@ export const getUserScope = (user: User): UserScopes => {
   }
 };
 
+export const sortAlphabetically = (a: User, b: User) => {
+  const nameA = a.name.toUpperCase(); // ignore upper and lowercase
```
kitajchuk, line 153: Prefer to lowercase and the comments are not necessary as it’s clear why we do this

src/utils/index.ts
```diff
@@ -148,11 +149,26 @@ export const getUserScope = (user: User): UserScopes => {
   }
 };
```
kitajchuk, line 151: This is should be called sortUsersAlpha or something or sortNamesAlpha so it’s use case is more clear

src/utils/use-scoped-redirect.ts
```diff
@@ -0,0 +1,47 @@
...
+  redirect: string,
+  user?: UserData,
+  message?: IMessage,
+  data?: Account | User | Application | Carrier | SpeechCredential
+) => {
+  const navigate = useNavigate();
+  const currentServiceProvider = useSelectState("currentServiceProvider");
+
+  useEffect(() => {
+    if (data && user?.access === 0 && data?.account_sid !== user?.account_sid) {
```
kitajchuk, line 26: Need to use the enum for all these checks please

kitajchuk (changes requested):
I’m assuming since @davehorton approved then the smoke and poke is functionally meeting requirements. However there are some changes to be made and some issues to be addressed here on the code review 👍

src/containers/internal/views/carriers/form.tsx
```diff
@@ -583,6 +606,21 @@ export const CarrierForm = ({
     }
   }
 
+  const checkOptions = () => {
+    if (user?.scope === USER_ACCOUNT) {
+      if (carrier && carrier.data?.account_sid) {
```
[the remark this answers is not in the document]
EgleHelms: this would work if carrier existing would be same value if account sid is there or not there. so this wont work the way I need it to if I make these change.

src/containers/internal/views/carriers/form.tsx
```diff
@@ -583,6 +606,21 @@ export const CarrierForm = ({
     }
   }
 
+  const checkOptions = () => {
```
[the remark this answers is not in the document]
EgleHelms: I actually moved it whole into utils since it is used two times.

src/containers/internal/views/phone-numbers/form.tsx
EgleHelms: @davehorton  asked me to redirect after the update for all resources to navigate back to the list of the resources.

src/containers/internal/views/users/form.tsx
EgleHelms: I just change it to normal if statement

src/containers/internal/views/users/index.tsx
EgleHelms: @davehorton could we return not only account_sid but also account_name and same for the service_provider_side and service_provider_name in the userJwt? 
Here I kind of hacked it in the frontend to list the names, but @kitajchuk thinks this should be returned byt the backend.

src/styles/_lists.scss
EgleHelms: I made a change where we list the resources without access to /edit page this it is seized to be a link. To make it similar looking I assigned a color.

src/containers/internal/navi/index.tsx
```diff
@@ -172,9 +174,55 @@ export const Navi = ({
...
+                  >
+                    <Item
+                      key={item.label}
+                      item={item}
+                      handleMenu={handleMenu}
+                    />
+                  </ScopedAccess>
+                );
+              }
+              if (item.label === "Accounts" && user?.scope === USER_ACCOUNT) {
```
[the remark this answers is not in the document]
EgleHelms: not sure how you see this old code, but this was resolved by merging your branch. I dont see this code anymore cannot even access it via history after rebasing.

src/containers/internal/views/phone-numbers/form.tsx
davehorton: yes the legacy webapp brings you back to the list after you add a resource, this branch leaves you on the resource you just added, which feels strange.  Update however is different in my mind, and you should stay on the resource after an update/save

davehorton: Having said that, I don't have a super-strong preference here and will defer to what you both think is best

src/containers/internal/views/users/index.tsx
davehorton: yes, will do

src/containers/internal/views/carriers/index.tsx
EgleHelms: I tried to address this, let me know what you think.

src/containers/internal/views/phone-numbers/form.tsx
EgleHelms: Actually, I check in the legacy app and seems we navigate back to the list in any case put or post. 
@kitajchuk let me know what you think, I will change it according to your prefference.

src/containers/internal/views/speech-services/edit.tsx
EgleHelms: totally right, not sure why this did not make it a problem in the console. Thanks.

src/containers/internal/views/users/index.tsx
davehorton: @EgleHelms the main branch of jambonz-api-server now encodes account name and sp name in jwt for account or sp level users

src/containers/internal/views/phone-numbers/form.tsx
davehorton: The main usability problem I have experienced is when you want to enter multiple new "things", whether that be users, phone numbers, applications.  You add the first one and after save are left on that newly-saved entity.  So you have an extra step to get back to the listview so you can click '+' again.  Or, you think you can add a second new entry direct from the screen you are now on, but actually any changes you make are not saved to a new entry but simply update the existing entry in place.

This is why I think the behavior post-Save for adding needs to be different than that for updating

kitajchuk: Ya that makes sense.

src/containers/internal/views/users/index.tsx
EgleHelms: @davehorton would it also be ok to add the same to user details in the get usersList? I think then it would cover everything, where we don't need to modify the data anywhere.

src/containers/internal/views/accounts/edit.tsx
```diff
@@ -19,12 +27,21 @@ export const EditAccount = () => {
...
+    user,
+    "You do not have access to this resource",
+    data
+  );
+
   /** Handle error toast at top level... */
   useEffect(() => {
     if (error) {
-      toastError(error.msg);
+      toastError(error.msg || "No access.");
```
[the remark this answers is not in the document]
EgleHelms: there is no need of them anymore, was left over from another solution I had tried.

src/containers/internal/views/phone-numbers/form.tsx
EgleHelms: OK, so now I am navigating to list after POST call, and navigating to the new resource /edit after put.
One exception is adding a predefined carrier, there I am navigating to /edit, since there might be changes one might want to do.

src/containers/internal/views/users/index.tsx
davehorton: this is done, on the latest main of jambonz-api-server

src/containers/internal/views/speech-services/status.tsx
```diff
@@ -17,10 +18,11 @@ export const CredentialStatus = ({
...
   showSummary = false,
 }: CredentialStatusProps) => {
+  const user = useSelectState("user");
+  const currentServiceProvider = useSelectState("currentServiceProvider");
+  const [url, setUrl] = useState("");
   const [testResult, testRefetch, testError] =
-    useServiceProviderData<CredentialTestResult>(
-      `SpeechCredentials/${cred.speech_credential_sid}/test`
-    );
+    useApiData<CredentialTestResult>(url);
```
EgleHelms, line 25: @kitajchuk I am not yet able to think of a good way for this. I tried to implement it how you asked with the useApiData, I didn't make any changes to that function just used a different URL to fetch the data. 
But in this current implementation, the url is not there yet, and there is an updated call made to a false URL. And then this not always resolving, like in the speech /edit page I used the same solution and it doesn't work. Maybe you can help a bit here.

src/containers/internal/views/carriers/index.tsx
EgleHelms: My solution is working for Carriers but not Speech, also it produced unwanted calls. Can you help with this?

src/containers/internal/views/users/index.tsx
EgleHelms: resolved

EgleHelms:
I am super confused, why does it need rebasing again, locally I needed to fetch and I got the latest rebased version of this branch, and then still needed to force-push my changes for whatever reason since I could not push it otherwise.
It is only me working on this project currently, and nothing was updated since the "IBM fix" merged to main. So why is there a need to do this again?

src/containers/internal/views/users/form.tsx
EgleHelms: This is changed: https://github.com/jambonz/jambonz-webapp/blob/311eb2d20973362526af5185dbb0e570c4e97794/src/containers/internal/views/users/form.tsx#L173
But does not appear here... same as the navi changes before for you...

src/containers/internal/views/speech-services/status.tsx
EgleHelms: @kitajchuk this is basically the last thing to be addressed, I think we should target to finish this asap (same structure in 3 places). I have addressed all other comments. We have to merge this before this week is over since I will be on holiday and unavailable.

kitajchuk:
@EgleHelms @davehorton I will take a look today to help fix the issues with the conditional react hooks.

EgleHelms:
@kitajchuk, I didn't see any changes pushed. ;/ hope you still have time today

kitajchuk:
> @kitajchuk, I didn't see any changes pushed. ;/ hope you still have time today

Yes sorry. Yesterday ended up too busy. I will prioritize it for today since I know you’re on a tight timeline here.

src/containers/internal/views/carriers/index.tsx
```diff
@@ -17,20 +18,43 @@ import {
...
 import { DeleteCarrier } from "./delete";
 
 import type { Account, Carrier, SipGateway, SmppGateway } from "src/api/types";
-import { Gateways } from "./gateways";
+import { Scope } from "src/store/types";
 
 export const Carriers = () => {
+  const user = useSelectState("user");
+  const currentServiceProvider = useSelectState("currentServiceProvider");
+  const [url, setUrl] = useState("");
```
kitajchuk, line 43: Looks like this is not being used here...

kitajchuk:
@EgleHelms @davehorton when i log in as a `service_provider` scope i still see the service provider selector in the side menu...my understanding is this is supposed to be hidden? or maybe we just want it to be readonly or something?

src/utils/index.ts
```diff
@@ -148,11 +159,53 @@ export const getUserScope = (user: User): UserScopes => {
...
 
+export const hasAccountAccess = (user: UserData, accountSid: string) => {
+  return user?.scope === USER_ACCOUNT && user.account_sid !== accountSid;
+};
+
+export const getUserAccounts = (user: UserData, accounts: Account[]) => {
+  return user?.scope === USER_ACCOUNT
+    ? accounts?.filter((acct) => acct.account_sid === user.account_sid)
```
kitajchuk, line 168: I believe I mentioned we should extract these utils, however I am now concerned about this. This kind of filtering should really be done at the API level. I should not be receiving data I shouldn't have access to and then relying on some frontend code to be there or not to filter it at the client level. This feels really backwards to me. @davehorton @EgleHelms thoughts on this?

kitajchuk: This is really unmanageable in its current incarnation. Having to copy this filtering code all over the frontend is brittle...

Screen Shot 2022-12-22 at 1 29 00 PM

src/containers/internal/views/carriers/index.tsx
EgleHelms: I am not sure I understand what you mean? It is used in the getUrlForCarriers where I stashed the condition. I have tried two ways: 
1. If I have the default url like here then the first render is wrong because the data is fetched from a wrong place 
2. if I use useApiData(url) the url being I conditioned in useEffect(running the getURL) the first render is then with undefined. 
So when you test it with the filter or go back and forth or refresh it is fetching from the wrong place.

In carriers what we need is to have it fetch form /accounts when filtering is applied with an accountSid and a /serviceProviders if all accounts are selected  in the filter. We also need to care where the /edit data is fetched if the users is account. And same with Speech providers.

src/containers/internal/views/speech-services/index.tsx
```diff
@@ -1,53 +1,83 @@
...
+  }, [accountSid, accounts, credentials]);
+
+  const filteredCredentials = useFilteredResults<SpeechCredential>(
+    filter,
+    credentialsFiltered
+  );
+
+  const getSpeechUrl = () => {
+    if (accountSid) {
+      setUrl(`Accounts/${user?.account_sid || accountSid}/SpeechCredentials`);
```
kitajchuk, line 57: This seems wrong...why am I checking `accountSid` but defaulting to `user?.account_sid`?

src/containers/internal/views/carriers/index.tsx
kitajchuk: PR coming though...

src/containers/internal/views/carriers/index.tsx
```diff
@@ -17,20 +18,43 @@ import {
...
+  const [carriers, refetch, error] = useApiData<Carrier[]>(
+    `ServiceProviders/${currentServiceProvider?.service_provider_sid}/VoipCarriers`
+  );
   const [accounts] = useServiceProviderData<Account[]>("Accounts");
   const [accountSid, setAccountSid] = useState("");
   const [filter, setFilter] = useState("");
 
+  const getUrlForCarriers = () => {
+    if ((user && user?.scope === USER_ACCOUNT) || accountSid) {
+      setUrl(`Accounts/${user?.account_sid || accountSid}/VoipCarriers`);
```
kitajchuk, line 54: But again this seems wrong to have the api url like this because it would always use the user account_sid...

src/containers/internal/views/carriers/index.tsx
kitajchuk: I'll post a PR and you and @davehorton can tell me if I've missed something. But some of this code feels strange to me. And I believe we're doing some things on the frontend that should be done on the backend (not this, but some other things).

kitajchuk:
@EgleHelms @davehorton when i log out as one user type and back in as another i often see this bad redirect and toast. seems to happen often if i log out of account or service_provider scope and then login as admin:

Screen Shot 2022-12-22 at 8 10 37 AM

src/containers/internal/views/speech-services/index.tsx
EgleHelms: In carriers might be unnecessary, but in the case of conditioning user.scope to account then makes sense.

src/containers/internal/views/carriers/index.tsx
EgleHelms: This is not the issue i am having, the url is wrong because of when it is rendered, condition is applied too late.

kitajchuk: It doesn't make sense because why am I selecting from an account filter if the value I'm selecting isn't being used to fetch the data? 👈 this doesn't make sense. In any case, the only account available in the filter is the one I'm scoped too...

kitajchuk: I know this isn't the issue you are having, PR for that is here: https://github.com/jambonz/jambonz-webapp/pull/168. But I'm still calling this out as it seems incorrect to me. Notes in PR code...

kitajchuk:
@davehorton @EgleHelms PR here for changes:
https://github.com/jambonz/jambonz-webapp/pull/168

But I have some questions/concerns about some things.

src/utils/index.ts
kitajchuk: @davehorton @EgleHelms This is one key concern of mine ☝️

EgleHelms: I think then when its that the user can only see his account I guess we could completely hide the filter. Or actually apply a different instance of filter component with only allowed things like All accounts and their account.

kitajchuk: Your second point is what actually already is happening because of this code. The data should be scoped on the backend, not on the frontend so we shouldn't be filtering api responses like this all over the place.

src/containers/internal/views/speech-services/index.tsx
kitajchuk: No this doesn't make sense for the same reason I mentioned in the other thread: why am I selecting an account in a filter if its value is never being used to fetch the data? 👈 that doesn't make sense.

davehorton:
> @EgleHelms @davehorton when i log in as a `service_provider` scope i still see the service provider selector in the side menu...my understanding is this is supposed to be hidden? or maybe we just want it to be readonly or something?

I'm ok either making it readonly or if the dropdown only has the one selection

src/utils/index.ts
davehorton: > I believe I mentioned we should extract these utils, however I am now concerned about this. This kind of filtering should really be done at the API level. I should not be receiving data I shouldn't have access to and then relying on some frontend code to be there or not to filter it at the client level. This feels really backwards to me. @davehorton @EgleHelms thoughts on this?

I agree.....are we getting data from the backend that is out of the user's scope?  If you can give me examples I can fix the back end.  Or maybe that is old code that is there from having to filter when using earlier back-end releases?

src/containers/internal/views/carriers/index.tsx
davehorton: Does this account for it?
- the path is `/ServiceProviders/{service_provider_sid}/VoipCarriers` to retrieve carriers defined at the SP level
- the path is `/Accounts/{account_sid}/VoipCarriers` to retrieve carriers defined at the Account level.

When the user is an SP user, we need to retrieve both - so we are going to retrieve account carriers but it will not be based on the user.account_sid..

src/utils/index.ts
kitajchuk: This was latest on main for api server. The frontend code here is filtering by user account sid. So it appears we fetch accounts not scoped to the user?

davehorton: Which API are we talking about here?  If it is the GET `/Users` that will return a list of users properly scoped for:
- an admin user (all users returned)
- an SP user (all users for that SP or users for child Accounts of that SP)
- an Account user (all users for that Account)

Is it that API or a different one?

kitajchuk: looks like all the calls are through `useServiceProviderData` hook which would be `/ServiceProviders/:service_provider_sid/Accounts`.

davehorton: I think a user with account scope should not be successful calling `/ServiceProviders/:service_provider_sid/Accounts` at all.  

As an account level user, you should get your own account by doing GET `Account/account_sid`.

So I am not sure why the front-end would need to use this service-provider level api in the first place.

I can look into the back end to see why this is even allowed

davehorton: OK I checked the code and I think the ServiceProvider retrieve functions were basically allowed by any user.  I have pushed a branch bugfix/fix-permissions which closes this loophole.

Now, in order to retrieve service provider data you must be either:
- an admin user, or
- a service provider user for the service provider entity you are querying

As an account level user, you will not be able to access this api.

Now, I imagine that initially this will break the webapp since given the description above it appears account level users have been making use of this API.  But, also per this discussion, I think that is wrong and leaks data to unauthorized users.

Please try testing with this branch of jambonz-api-server and see if we can work out what needs to change as a result on the webapp.

kitajchuk: That sounds correct. So I think for account level users we should omit the account filters on all list views since you should only see resources for your account. And for forms with account selectors we could do the same or just let the selectors only have the one account. This poses an issue because the web app assumes to be fetching accounts plural in all these cases so switching to an api that returns a single object breaks the model there… 🤔

davehorton: hmm, in thinking about this some more, I realize that we do need to show account users the carriers and speech credentials that have been defined at their parent service provider level -- we show them, but do not let them update.  So I think my changes to forbid this will create a problem.  I need to think on this a bit more, I will be commiting some more changes to that jambonz-api-server branch tomorrow to allow this

kitajchuk: Are all endpoints based off ServiceProvider blocked for account level users? If so, this will break across nearly every view for account level users. We would need to have a conditional api url for all resources in that case. That’s not idea if it’s the case. Can you confirm?

src/containers/internal/views/speech-services/index.tsx
EgleHelms: resolved with your pr.

EgleHelms:
@kitajchuk @davehorton 

I have applied these changes:
1) removed comments that were merged from @kitajchuk PR questioning some unneeded conditions - seems those conditions were not needed.
2) made serviceProvider selector read-only by disabling it
3) removed Account filters from the Account user views where they should not see other account information but had to then conditionally assign an accountSid of the user for lists to be correctly fetched. This way we have gotten rid of some of the getUserAccounts filtering. ( This point was applied to Users, Applications, Recent Calls, Alerts, Phone Numbers views) - I was not sure if there is a need to do this for the MS teams list (let me know)
4) used parseJwt to address the issue of the "user" state is wrong when logging in as a different user and using rerouting. I could not see any other solution for this, @kitajchuk is there one?

The current state of the main branch on jambonz-api-server seems to still allow everything we need on the frontend so far, but there are still a getUserAccounts cases.

I would suggest then the backend allows fetching the accounts from /ServiceProvider but for Account scoped user gets one account in an array, this way we can remove the filtering for the account filter and have the structure on the frontend work correctly, I think.

EgleHelms:
One more thing, how do we resolve conversations from commits that don't exist anymore?
Two conversations I seem to not be able to get to anymore.

src/utils/index.ts
davehorton: Yes, I think I will have to allow account users to call the ServiceProvider GET api, but filter what is returned to only show what that account level user is allowed.  I will work on a change for that today on the new branch of api-server

davehorton:
> @davehorton @kitajchuk I would suggest then the backend allows fetching the accounts from /ServiceProvider but for Account scoped user gets one account in an array, this way we can remove the filtering for the account filter and have the structure on the frontend work correctly, I think.

Agree,

kitajchuk:
> > @davehorton @kitajchuk I would suggest then the backend allows fetching the accounts from /ServiceProvider but for Account scoped user gets one account in an array, this way we can remove the filtering for the account filter and have the structure on the frontend work correctly, I think.
> 
> Agree,

Yes that would be ideal so we don’t need to worry about a difference in the data structure 👍

davehorton:
> > > @davehorton @kitajchuk I would suggest then the backend allows fetching the accounts from /ServiceProvider but for Account scoped user gets one account in an array, this way we can remove the filtering for the account filter and have the structure on the frontend work correctly, I think.
> > 
> > 
> > Agree,
> 
> Yes that would be ideal so we don’t need to worry about a difference in the data structure 👍

I have pushed a change to the bugfix/fix-permissions branch of jambonz-api-server that allows account users to access the /ServiceProviders api but filters the data returned so as to not give them visibility into other account info or service provider info that should be off limits.  Next step I think is for @EgleHelms to push a change where she has removed the filtering on the webapp side, as it should now not be needed.

davehorton:
I pushed a change (above) to remove that account filter as it is no longer needed.  I also merged the accompanying change on the jambonz-api-server to the main branch.

src/utils/index.ts
```diff
@@ -163,9 +163,7 @@ export const hasAccountAccess = (user: UserData, accountSid: string) => {
 };
 
 export const getUserAccounts = (user: UserData, accounts: Account[]) => {
-  return user?.scope === USER_ACCOUNT
-    ? accounts?.filter((acct) => acct.account_sid === user.account_sid)
-    : accounts;
+  return accounts;
```
kitajchuk, line 166: @davehorton we should actually remove this entire function and just pass accounts as props directly like we did before. A function that just returns what you pass in is not needed 🤔

kitajchuk: I have some time this AM so I can do a final code-pass on this 👍

kitajchuk:
@davehorton I get a 403 forbidden when attempting to create a new user with an email that exists for another user already. This isn't a very useful response because the web app can only toast generically like "something went wrong". It would be good to leverage our error responses that have the custom `msg` field so we can be more clear about these issues 🤔

src/utils/index.ts
davehorton: that would be great

kitajchuk:
@davehorton Looks like as an `account` user I am told I don't have proper permissions to create a carrier from a predefined option. However, when I go back to the list view and switch the account filter I see the carriers were saved on the backend still...

Screen Shot 2022-12-26 at 8 29 55 AM

Screen Shot 2022-12-26 at 8 30 50 AM

kitajchuk:
@davehorton Also on the users table view I am able to filter by all scopes even when I'm logged in as service_provider or account level scope which seems like my scope filter should match based on my create capabilities 🤔

kitajchuk:
@davehorton @EgleHelms I pushed a couple commits to clean a few things up. I have a couple open questions above so once those are clarified/addressed things look good on my end 👍

davehorton:
> @davehorton Looks like as an `account` user I am told I don't have proper permissions to create a carrier from a predefined option. However, when I go back to the list view and switch the account filter I see the carriers were saved on the backend still...

Hmm..the issue here is that we have created a Carrier at the Account level, yet then when we want to update it we do a PUT to `/ServiceProviders/:sid/VoipCarriers/:voip_carrier_sid`.  

I think the webapp should be doing a PUT to `/Accounts/:sid/VoipCarriers/:voip_carrier_sid` instead, as that is more appropriate.

The only problem is that the back-end has not implemented that second API yet.  But I think the right thing is for me to implement that, and then for the front end to use that api when updating an account Carrier.

Do you agree @kitajchuk ?

davehorton:
I've pushed a change to jambonz-api-server main branch to add the PUT `/Accounts/:account_sid/VoipCarriers/:carrier_sid`.

I suppose, though, if it is easier I can change the backend to allow an account level user to access that specific service provider api.

kitajchuk:
> I've pushed a change to jambonz-api-server main branch to add the PUT `/Accounts/:account_sid/VoipCarriers/:carrier_sid`.
> 
> I suppose, though, if it is easier I can change the backend to allow an account level user to access that specific service provider api.

I think using this api is better as it is in line with our other account specific apis. I can make the change here

kitajchuk:
@davehorton The new account-level PUT api for carriers is implemented with this commit:
https://github.com/jambonz/jambonz-webapp/pull/159/commits/ccca4da4c2eaaf38ec6b3270d66f677a0430be08

davehorton:
> @davehorton The new account-level PUT api for carriers is implemented with this commit: [9ae4a0b](https://github.com/jambonz/jambonz-webapp/commit/9ae4a0b5acb3c11ab6a9c6ab96964ec01af48df8)

Looks good in my testing!

davehorton:
> @davehorton Also on the users table view I am able to filter by all scopes even when I'm logged in as service_provider or account level scope which seems like my scope filter should match based on my create capabilities 🤔

I agree that the selections in the Scope dropdown should be filtered:
- admin user sees all
- sp user sees only service provider and account
- account user sees only account -- or, better, the dropdown is hidden entirely since it has only the one selection

davehorton:
> @davehorton I get a 403 forbidden when attempting to create a new user with an email that exists for another user already. This isn't a very useful response because the web app can only toast generically like "something went wrong". It would be good to leverage our error responses that have the custom `msg` field so we can be more clear about these issues 🤔

I pushed a commit to jambonz-api-server to return a json payload with `msg` so now this does show the toast.  Thanks for the idea!

davehorton (approved):

kitajchuk:
> > @davehorton Also on the users table view I am able to filter by all scopes even when I'm logged in as service_provider or account level scope which seems like my scope filter should match based on my create capabilities 🤔
> 
> I agree that the selections in the Scope dropdown should be filtered:
> 
> * admin user sees all
> * sp user sees only service provider and account
> * account user sees only account -- or, better, the dropdown is hidden entirely since it has only the one selection

I think currently the filter is already hidden for account users so would just need to properly filter what scopes are present for sp since admins see all.

EgleHelms:
Wow, nice work you guys are lightning fast. Thanks for pushing this to the finish line. 

@kitajchuk I cannot get to the unresolved conversations anymore to resolve them probably because rebasing it is 404 when i click on the two conversations, I feel like I answered those too that time. Can you let me know what are the other questions? I feel like I cannot find anything else unanswered.

davehorton:
@kitajchuk is this ready to merge from your perspective?

kitajchuk:
> @kitajchuk is this ready to merge from your perspective?

I think maybe the account filter scoping hasn’t been implemented?

EgleHelms:
@davehorton @kitajchuk I pushed a small commit to add the filtering for the User list according to user scope.
Hope this is now ready to be merged.

kitajchuk:
> @davehorton @kitajchuk I pushed a small commit to add the filtering for the User list according to user scope.
> Hope this is now ready to be merged.

Ya that commit looks good 👍. I’d say if @davehorton has smoke and poked this and it looks good then you can squash and merge this PR 🎉

davehorton (approved):

PR #160, opened 2022-12-01
Add IBM as speech provider

EgleHelms:
- Added to Speech form
- Added to Application form

Was not sure about the util this time either. Since I didnt need to add to make the conditionals bigger.

### DIFF
--- a/src/vendor/speech-recognizer/ibm-speech-recognizer-lang.ts
+++ b/src/vendor/speech-recognizer/ibm-speech-recognizer-lang.ts
@@ -1,0 +1,130 @@
+import type { Language } from "../types";
+
+export const languages: Language[] = [
+  {
+    name: "Arabic (Modern Standard) - Telephony",
+    code: "ar-MS_Telephony",
+  },
+  {
+    name: "Chinese (Mandarin) - Telephony",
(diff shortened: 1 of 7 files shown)


davehorton:
So the STT language choices should only include the telephony ones.  The label should not include "Telephony" but the value should ie value=en-US_Telephony

so instead of 
```
English (United States) - Telephony
English (United States) - MultiMedia
```

it should just be
```
English (United States)
```
but the value sent should be en-US_Telephony

src/vendor/speech-recognizer/ibm-speech-recognizer-lang.ts
```diff
@@ -0,0 +1,130 @@
...
+  {
+    name: "English (all supported dialects) - Telephony",
+    code: "en-WW_Medical_Telephony",
+  },
+  {
+    name: "English (Australian) - Telephony",
+    code: "en-AU_Telephony",
+  },
+  {
+    name: "English (Australian) - Multimedia",
```
davehorton, line 33: only return the Telephony options, not Multimedia (and remove "- Telephony" from the name

davehorton (changes requested):

davehorton (approved):

PR #163, opened 2022-12-08
Refactor how we manage scope for navi

kitajchuk:


davehorton (approved):

EgleHelms:
@kitajchuk yeah, this is definitely a better way. Thanks a lot.

PR #164, opened 2022-12-09
Fix useScopedRedirect hook

kitajchuk:
Also includes updates to the user UI:

Screen Shot 2022-12-09 at 8 35 53 AM
Screen Shot 2022-12-09 at 8 36 00 AM
Screen Shot 2022-12-09 at 8 36 07 AM


kitajchuk:
@davehorton @EgleHelms this branch can be merged into the feature scope branch with fixes for the redirect hook 👍

PR #169, opened 2023-01-06
Scope optional limits

EgleHelms:
@davehorton @kitajchuk This was the code that was lost somehow during the scoping. 
As agreed, we wanted the new call limits under the env VITE_APP_ENABLE_ACCOUNT_LIMITS_ALL to be disabled for non-admin users, and the MAX limits hidden completely.

Account User view with env VITE_APP_ENABLE_ACCOUNT_LIMITS_ALL
image

Admin user view env VITE_APP_ENABLE_ACCOUNT_LIMITS_ALL
image

There is still a question on where we should handle the deletion of the limits set the first time that are in contradiction with the new limits set (e.g. we set session limits, then we decide to change to minutes limits the session limits are not deleted all limits we send more limits and all exist in the backend - so the old ones should be deleted). @davehorton @kitajchuk what do you think should we do this in the frontend?

### DIFF
--- a/src/api/constants.ts
+++ b/src/api/constants.ts
@@ -152,17 +152,17 @@ export const LIMITS: LimitField[] = [
   },
 ];
 
-export const LIMIT_MINS = "minutes";
-export const LIMIT_SESS = "sessions";
+export const LIMIT_MIN = "minute";
+export const LIMIT_SESS = "session";
 
 export const LIMIT_UNITS: LimitUnitOption[] = [
   {
-    name: "Sessions",
+    name: "Session",
     value: LIMIT_SESS,
   },
   {
-    name: "Minutes",
-    value: LIMIT_MINS,
+    name: "Minute",
+    value: LIMIT_MIN,
   },
 ];
 
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -327,6 +327,7 @@ export const postServiceProviderLimit = (
   sid: string,
   payload: Partial<Limit>
 ) => {
+  console.log(payload);
   return postFetch<SidResponse, Partial<Limit>>(
     `${API_SERVICE_PROVIDERS}/${sid}/Limits`,
     payload
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -16,7 +16,7 @@ export type LimitCategories =
   | "voice_call_minutes"
   | "voice_call_session_license";
 
-export type LimitUnit = "Sessions" | "Minutes";
+export type LimitUnit = "Session" | "Minute";
 
 export interface LimitUnitOption {
   name: LimitUnit;
--- a/src/components/forms/local-limits.tsx
+++ b/src/components/forms/local-limits.tsx
@@ -1,10 +1,19 @@
 import React, { useEffect, useRef, useState } from "react";
 
-import { LIMITS, LIMIT_MINS, LIMIT_SESS, LIMIT_UNITS } from "src/api/constants";
+import {
+  LIMITS,
+  LIMIT_MINS,
+  LIMIT_SESS,
+  LIMIT_UNITS,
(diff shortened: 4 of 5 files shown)


davehorton:
> There is still a question on where we should handle the deletion of the limits set the first time that are in contradiction with the new limits set (e.g. we set session limits, then we decide to change to minutes limits the session limits are not deleted all limits we send more limits and all exist in the backend - so the old ones should be deleted). @davehorton @kitajchuk what do you think should we do this in the frontend?

Yes, I think this should be done in the front end (explicitly deleting the "old" limits)

EgleHelms:
@kitajchuk @davehorton this is ready for review.

Changes:
- Service provider Limits disabled when the env is enabled
- No limits or uinits are available to change for non-admin users (all are disabled)
- Max limits both minute or session are hidden for service_provider and account users
- quantity is set to "" when unit is changed to initiate deletion

davehorton:
@EgleHelms  Perhaps I am doing something wrong but in testing this I am seeing an error in the console and I am seeing neither the Units nor the Licensed minutes on the Account screen.  I am logged in as admin user and my .env.local has `VITE_APP_ENABLE_ACCOUNT_LIMITS_ALL=true`

image

kitajchuk:
> @EgleHelms Perhaps I am doing something wrong but in testing this I am seeing an error in the console and I am seeing neither the Units nor the Licensed minutes on the Account screen. I am logged in as admin user and my .env.local has `VITE_APP_ENABLE_ACCOUNT_LIMITS_ALL=true`
> 
> image

That’s a react key error—should be easily fixable but I’m curious why it would show up suddenly 🤔

davehorton:
let me retest and make sure I am not doing anything stupid

davehorton:
yeah I retested to make sure I am testing this PR and it does look like a bug here

davehorton:
I put in some logging and I think the issue is that "category" is used as the key and it is not unique
```
filtered limits: [{"label":"Max calls","category":"voice_call_session"},{"label":"Licensed calls","category":"voice_call_session_license"}]
```

EgleHelms:
hmm, I wonder, why would it not be unique? the four categories have different category names. When I was fixing, it was not popping up.
@kitajchuk would you just add a random id to the key to create a unique one? something like key={`${category}-${Math.floor + Math.random() * 100000 )}` It seems to solve the problem, but I am not sure if this is the right solution or is this a dumb solution?

Also, this error, was not preventing me from viewing the fields. They also work to submit the limits too.

image

EgleHelms:
@kitajchuk @davehorton  sorry this is a bit dumb issue I left in, the issue was that I have added another Component <ScopedAccess> that one needs to have a key too since it is also within the map function. So I added the key and the issue is gone. Though the error in the console did not prevent me from viewing the fields and it and somehow is not always there in the console, so I have missed it.

davehorton (approved):

kitajchuk:
> @kitajchuk @davehorton sorry this is a bit dumb issue I left in, the issue was that I have added another Component `<ScopedAccess>` that one needs to have a key too since it is also within the map function. So I added the key and the issue is gone. Though the error in the console did not prevent me from viewing the fields and it and somehow is not always there in the console, so I have missed it.

This code got merged with duplicate keys explicitly coded into the map—this is not good. Can we please fix that. Should be easy—even just using string template. Or remove the key on the fragment as it looks like that is not needed—only the direct component returned at the root of the map callback needs a key. 

https://reactjs.org/docs/lists-and-keys.html#keys

kitajchuk:
> My bad for merging that, @EgleHelms maybe just commit this change on the open PR for date fix (which also needs to be rebased to main)?

Just remove the fragment as it’s no longer needed since you added the ScopedAccess wrapper.

Issue #170, opened 2023-01-08
latest webapp (main branch) only works in Chrome with an incognito window


davehorton:
Note sure why @kitajchuk @EgleHelms  but when I run the latest webapp on a fully deployed server, when I go to the URL which is hosting jambonz I am immediately brought to a blank screen and the url is http:://<url>/internal/accounts instead of the login screen:

![image](https://user-images.githubusercontent.com/1884973/211177551-c8f8277b-2552-48c0-8e4b-1c195af9a43f.png)

On the other hand, if I am in a Chrome incognito window, it works
![image](https://user-images.githubusercontent.com/1884973/211177605-602dbe6d-4a36-46a0-9136-96b0a406afb7.png)

I can provide you access to the jambonz.one server to investigate if needed.

kitajchuk:
> Note sure why @kitajchuk @EgleHelms but when I run the latest webapp on a fully deployed server, when I go to the URL which is hosting jambonz I am immediately brought to a blank screen and the url is http::///internal/accounts instead of the login screen:
> 
> ![image](https://user-images.githubusercontent.com/1884973/211177551-c8f8277b-2552-48c0-8e4b-1c195af9a43f.png)
> 
> On the other hand, if I am in a Chrome incognito window, it works ![image](https://user-images.githubusercontent.com/1884973/211177605-602dbe6d-4a36-46a0-9136-96b0a406afb7.png)
> 
> I can provide you access to the jambonz.one server to investigate if needed.

Did you try clearing your local and session storage?

kitajchuk:
> > Note sure why @kitajchuk @EgleHelms but when I run the latest webapp on a fully deployed server, when I go to the URL which is hosting jambonz I am immediately brought to a blank screen and the url is http::///internal/accounts instead of the login screen:
> > ![image](https://user-images.githubusercontent.com/1884973/211177551-c8f8277b-2552-48c0-8e4b-1c195af9a43f.png)
> > On the other hand, if I am in a Chrome incognito window, it works ![image](https://user-images.githubusercontent.com/1884973/211177605-602dbe6d-4a36-46a0-9136-96b0a406afb7.png)
> > I can provide you access to the jambonz.one server to investigate if needed.
> 
> Did you try clearing your local and session storage?

That’s probably it because it worked for me and I’ve never been to that url before.

davehorton:
[...]

PR #172, opened 2023-01-08
Issue #170: Fix bug with bad user reference in internal accounts page

kitajchuk:


kitajchuk:
@davehorton I sure enough did replicate the bug locally by adding a token in my storage that was NOT a JWT. But, the real issue here was using a bad reference to the `user` in a component. We need to always pull the current user from the `store` so that is what the root cause of the bug was.

davehorton (approved):

EgleHelms:
@kitajchuk this fix makes the bug of logging out as an account user and then logging in as admin throw this error since the state is not updated in time. I thought this was actually agreed on in the Scoping PR we merged.

image

EgleHelms:
Maybe a fix here could be not redirecting to something that is accessible to all. Like "Applications"?

davehorton:
"state not updated in time" sounds like a strange race condition we ought to fix.  Once you are logged out, why is there any state at all kept?

kitajchuk:
@davehorton @EgleHelms This PR fixes the issue of logging out and switching users: https://github.com/jambonz/jambonz-webapp/pull/174

PR #176, opened 2023-01-13
Date fix for alerts and recent calls

EgleHelms:
"Invalid date" was displayed instead of a title for Alerts and recent calls

image

The date format received:

{
  "call": {
    "attempted_at": "2022-12-28T18:43:53.874Z",
    "account_sid": "xxx",
    "answered": true,
    "answered_at": "2022-12-28T18:43:56.841Z",
    "application_sid": "xxx",
    "call_sid": "xxx",
    "direction": "inbound",
    "duration": 17,
    "from": "+xxxxxxx",
    "host": "undefined",
    "remote_host": "54.172.60.3",
    "service_provider_sid": "2708b1b3-2736-40ea-b502-c53d8396247f",
    "sip_callid": "xxxx@0.0.0.0",
    "sip_status": 200,
    "terminated_at": "2022-12-28T18:44:14.226Z",
    "termination_reason": "called party hungup",
    "to": "+15082069511",
    "trace_id": "xxx",
    "trunk": "Twilio"
  }
}

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -530,19 +530,19 @@ export const getRecentCalls = (sid: string, query: Partial<CallQuery>) => {
   );
 };
 
-export const getRecentCall = (sid: string, callSid: string) => {
+export const getRecentCall = (sid: string, sipCallId: string) => {
   return getFetch<TotalResponse>(
     import.meta.env.DEV
-      ? `${DEV_BASE_URL}/Accounts/${sid}/RecentCalls/${callSid}`
-      : `${API_ACCOUNTS}/${sid}/RecentCalls/${callSid}`
+      ? `${DEV_BASE_URL}/Accounts/${sid}/RecentCalls/${sipCallId}`
+      : `${API_ACCOUNTS}/${sid}/RecentCalls/${sipCallId}`
   );
 };
 
-export const getPcap = (sid: string, callSid: string) => {
+export const getPcap = (sid: string, sipCallId: string) => {
   return getBlob(
     import.meta.env.DEV
-      ? `${DEV_BASE_URL}/Accounts/${sid}/RecentCalls/${callSid}/pcap`
-      : `${API_ACCOUNTS}/${sid}/RecentCalls/${callSid}/pcap`
+      ? `${DEV_BASE_URL}/Accounts/${sid}/RecentCalls/${sipCallId}/pcap`
+      : `${API_ACCOUNTS}/${sid}/RecentCalls/${sipCallId}/pcap`
   );
 };
 
--- a/src/components/forms/local-limits.tsx
+++ b/src/components/forms/local-limits.tsx
@@ -110,7 +110,7 @@ export const LocalLimits = ({
                 Scope.account
               }
             >
-              <React.Fragment key={category}>
+              <React.Fragment>
                 <label htmlFor={category}>{label}</label>
                 <input
                   ref={(el: HTMLInputElement) => {
--- a/src/containers/internal/views/alerts/index.tsx
+++ b/src/containers/internal/views/alerts/index.tsx
@@ -101,9 +101,7 @@ export const Alerts = () => {
                 <div className="item__info">
                   <div className="item__title txt--jam">
                     <strong>
-                      {dayjs
-                        .unix(alert.time / 1000)
-                        .format("YYYY MM.DD hh:mm a")}
+                      {dayjs(alert.time).format("YYYY MM.DD hh:mm a")}
                     </strong>
                   </div>
                   <div className="item__meta">
--- a/src/containers/internal/views/recent-calls/details.tsx
+++ b/src/containers/internal/views/recent-calls/details.tsx
@@ -28,9 +28,7 @@ export const DetailsItem = ({ call }: DetailsItemProps) => {
           <div className="item__info">
             <div className="item__title">
               <strong>
-                {dayjs
-                  .unix(call.attempted_at / 1000)
-                  .format("YYYY MM.DD hh:mm a")}
+                {dayjs(call.attempted_at).format("YYYY MM.DD hh:mm a")}
               </strong>
               <span className="i txt--dark">
                 {call.direction === "inbound" ? (
--- a/src/containers/internal/views/recent-calls/details.tsx
+++ b/src/containers/internal/views/recent-calls/details.tsx
@@ -60,7 +60,11 @@ export const DetailsItem = ({ call }: DetailsItemProps) => {
             {Object.keys(call).map((key) => (
               <React.Fragment key={key}>
                 <div>{key}:</div>
-                <div>{call[key as keyof typeof call].toString()}</div>
+                <div>
+                  {call[key as keyof typeof call]
+                    ? call[key as keyof typeof call].toString()
+                    : "null"}
+                </div>
               </React.Fragment>
             ))}
           </div>
--- a/src/containers/internal/views/recent-calls/pcap.tsx
+++ b/src/containers/internal/views/recent-calls/pcap.tsx
@@ -13,10 +13,10 @@ export const PcapButton = ({ call }: PcapButtonProps) => {
   const [pcap, setPcap] = useState<Pcap>();
 
   useEffect(() => {
-    getRecentCall(call.account_sid, call.call_sid)
+    getRecentCall(call.account_sid, call.sip_callid)
       .then(({ json }) => {
         if (json.total > 0) {
-          getPcap(call.account_sid, call.call_sid)
+          getPcap(call.account_sid, call.sip_callid)
             .then(({ blob }) => {
               if (blob) {
                 setPcap({


kitajchuk:
> I get a new and different error now when testing this. I've installed it on jambonz.one and when I initially click on recent calls it show no calls for today, which is correct. If I think click on "last 7 days" I get an error in the console and a blank window
> 
> image
> 
> I verified that the API call to the server was fine: image

Is this unique to this branch though?

kitajchuk:
> > I get a new and different error now when testing this. I've installed it on jambonz.one and when I initially click on recent calls it show no calls for today, which is correct. If I think click on "last 7 days" I get an error in the console and a blank window
> > image
> > I verified that the API call to the server was fine: image
> 
> Is this unique to this branch though?

Oh I see this is a different PR lol. I’m wondering if this line is the culprit:

https://github.com/jambonz/jambonz-webapp/blob/55c1b26ddca72eb35e795b22fcadc84265304f17/src/containers/internal/views/recent-calls/details.tsx#L65

Maybe not but it’s the only thing calling toString in the render by us. Otherwise maybe the date stuff. Seems like some value can be null and we’re not expecting it…

kitajchuk:
> > > I get a new and different error now when testing this. I've installed it on jambonz.one and when I initially click on recent calls it show no calls for today, which is correct. If I think click on "last 7 days" I get an error in the console and a blank window
> > > image
> > > I verified that the API call to the server was fine: image
> > 
> > 
> > Is this unique to this branch though?
> 
> Oh I see this is a different PR lol. I’m wondering if this line is the culprit:
> 
> https://github.com/jambonz/jambonz-webapp/blob/55c1b26ddca72eb35e795b22fcadc84265304f17/src/containers/internal/views/recent-calls/details.tsx#L65
> 
> Maybe not but it’s the only thing calling toString in the render by us. Otherwise maybe the date stuff. Seems like some value can be null and we’re not expecting it…

Either way this line is sus. There should be some checking on the value since we’re just mapping the object keys here and presumably the value can be null in some cases for some of these properties 🤔

davehorton:
OK, now the problem manifests differently.  See below, the list of recent calls now populates, but when I click on a call to get the detail the webapp makes a GET to /Accounts/:sid/RecentCalls/null -- ie where the callSid should be it is null.

@EgleHelms you should be able to test against jambonz.one on the backend with your webapp running locally by configuring your .env:
```
VITE_API_BASE_URL=http://jambonz.one/api/v1
```
image

davehorton:
also @EgleHelms  once you fix this please see if a similar fix is required for the Alerts panel since they are pretty similar

EgleHelms:
> OK, now the problem manifests differently. See below, the list of recent calls now populates, but when I click on a call to get the detail the webapp makes a GET to /Accounts/:sid/RecentCalls/null -- ie where the callSid should be it is null.
> 
> @EgleHelms you should be able to test against jambonz.one on the backend with your webapp running locally by configuring your .env:
> 
> ```
> VITE_API_BASE_URL=http://jambonz.one/api/v1
> ```
> 
> image

@davehorton I mean but this is the data I am getting, the sid there is indeed null. I am not sure why, because the calls were failed maybe?

This is for example one of the calls data: 


```
{
  "attempted_at": "2023-01-12T19:33:17.771Z",
  "account_sid": "9351f46a-678c-43f5-b8a6-d4eb58d131af",
  "answered": false,
  "answered_at": "1970-01-01T00:00:00.000Z",
  "application_sid": null,
  "call_sid": null,
  "direction": "inbound",
  "duration": 0,
  "from": "+15083084809",
  "host": "52.201.207.0",
  "remote_host": "54.172.60.1",
  "service_provider_sid": "2708b1b3-2736-40ea-b502-c53d8396247f",
  "sip_callid": "1f48846ed31708b12d8247c8138bdf7b@0.0.0.0",
  "sip_status": 500,
  "terminated_at": "2023-01-12T19:33:17.786Z",
  "termination_reason": "failed",
  "to": "+15086908019",
  "trace_id": null,
  "trunk": "Twilio"
}
```

davehorton:
Hmm, I will have a look.  The sip status was a 500, which does imply some sort of application error but I think there should always be a callSid.  Let me look at the back-end code and also make some new calls to see how those display

EgleHelms:
> > Hmm, I will have a look. The sip status was a 500, which does imply some sort of application error but I think there should always be a callSid. Let me look at the back-end code and also make some new calls to see how those display
> 
> Seems like yes the callSid should always be present but I would still recommend adding checks to that object values map so the Frontend code is more resilient, even in such edge cases as this.

@kitajchuk  I am not sure what you mean by resilient here. Do you mean like adding a strict type to the call object and its values? If we would add the type and the type is not exact, it will also fail. Then we need some error about malformed data received that we will not be able to display the details then or at least not the values that are malformed? Can you please expand on your idea of resilience?

kitajchuk:
> > > Hmm, I will have a look. The sip status was a 500, which does imply some sort of application error but I think there should always be a callSid. Let me look at the back-end code and also make some new calls to see how those display
> > 
> > 
> > Seems like yes the callSid should always be present but I would still recommend adding checks to that object values map so the Frontend code is more resilient, even in such edge cases as this.
> 
> @kitajchuk I am not sure what you mean by resilient here. Do you mean like adding a strict type to the call object and its values? If we would add the type and the type is not exact, it will also fail. Then we need some error about malformed data received that we will not be able to display the details then or at least not the values that are malformed? Can you please expand on your idea of resilience?

Hmm, maybe hadn’t thought it through to that level yet. I was just thinking at the very least so the code doesn’t throw a runtime error.

davehorton:
OK, the problem is stranger still.  This time, I made a successful call on jambonz.one so that I could have one to look at.  

When I click on Recent calls and expand that call, all looks pretty good at first glance:

image

The details are all there and are all correct.  Notice particularly that:
- call_sid is 3417d0d8-b084-4c3d-b97d-92a903db94a8
- sip_call_id is 1b5fe364aace211bc01f59d996dd9839@0.0.0.0

Now however, notice that the "pcap" button is not showing, and it should.  
This indicates the fetch to retrieve pcap data from homer failed for some reason.  Let's look at that API call:

image

Notice that the URL is putting the call_sid in the last param where instead the sip_call_id should go,
ie it sends this
```
 http://jambonz.one/api/v1/Accounts/9351f46a-678c-43f5-b8a6-d4eb58d131af/RecentCalls/3417d0d8-b084-4c3d-b97d-92a903db94a8
```
whereas it should be sending this
```
 http://jambonz.one/api/v1/Accounts/9351f46a-678c-43f5-b8a6-d4eb58d131af/RecentCalls/1b5fe364aace211bc01f59d996dd9839@0.0.0.0
```
Because of this, we try to find a call in homer with a bogus call-id and it fails.

davehorton:
Yeah from looking at the code it seems like there is a misunderstanding about that API.  It needs to retrieve by sip_call_id not call_sid because we are getting the data from homer where it is tracked/indexed by SIP Call-ID

kitajchuk:
> Yeah from looking at the code it seems like there is a misunderstanding about that API. It needs to retrieve by sip_call_id not call_sid because we are getting the data from homer where it is tracked/indexed by SIP Call-ID

Just need to change it here:
https://github.com/jambonz/jambonz-webapp/blob/def7743b9eb55446999807df05249de766e10f66/src/containers/internal/views/recent-calls/pcap.tsx#L19

davehorton:
probably should change it [here](https://github.com/jambonz/jambonz-webapp/blob/def7743b9eb55446999807df05249de766e10f66/src/api/index.ts#L533) as well to make clear:

davehorton:
also, of course, we need this to be rebased to main

kitajchuk:
> probably should change it [here](https://github.com/jambonz/jambonz-webapp/blob/def7743b9eb55446999807df05249de766e10f66/src/api/index.ts#L533) as well to make clear:

Ya @EgleHelms we should rename that argument to be sipCallId 👍 and actually pass the sip_call_id in where we call that.

EgleHelms:
> > > > Hmm, I will have a look. The sip status was a 500, which does imply some sort of application error but I think there should always be a callSid. Let me look at the back-end code and also make some new calls to see how those display
> > > 
> > > 
> > > Seems like yes the callSid should always be present but I would still recommend adding checks to that object values map so the Frontend code is more resilient, even in such edge cases as this.
> > 
> > 
> > @kitajchuk I am not sure what you mean by resilient here. Do you mean like adding a strict type to the call object and its values? If we would add the type and the type is not exact, it will also fail. Then we need some error about malformed data received that we will not be able to display the details then or at least not the values that are malformed. Can you please expand on your idea of resilience?
> 
> Hmm, maybe hadn’t thought it through to that level yet. I was just thinking at the very least so the code doesn’t throw a runtime error.

@kitajchuk I still am not sure what you mean. Just let me know or feel free to push anything into the branch if you have something specific in mind that is too hard to explain. I can only think of validation or a fallback on error... maybe and ErrorBoundry component or so.

EgleHelms:
When I am testing now before or after the changes in seems to not matter "Recent calls" and "Alerts" I keep getting another kind of error that was not there before.
`Unexpected token '<', "<!DOCTYPE "... is not valid JSON`

EgleHelms:
When I am testing connecting to jambonz.one with env var I just get a `Unexpected token '<', "<!DOCTYPE "... is not valid JSON` error when fetching data.

davehorton:
I've tested the latest commit and it now works on jambonz.one -- shows the pcap button and downloads the pcap.
@kitajchuk do you want to have a final look and merge if it looks ok?

davehorton (approved):

EgleHelms:
@davehorton @kitajchuk please let's merge this. If no new comments came back.

PR #179, opened 2023-01-19
Always attach service_provider_sid to a carrier

EgleHelms:
Service provider sid was not attached when an account level carrier was created, thus Phone number fetch of carriers did not contain account-related carriers.

This now allows the Account user to create account level Carrier and then add a Phone number.

I still think that we should restrict /SP API to return only relevant account carriers. Now if there are more accounts all will be fetched.

Related PR was created to limit /SP API to scope for both VoipCarriers and Speech Creds
https://github.com/jambonz/jambonz-api-server/pull/96

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 VITE_API_BASE_URL=http://127.0.0.1:3000/v1
-VITE_DEV_BASE_URL=http://127.0.0.1:3002/api
+VITE_DEV_BASE_URL=http://127.0.0.1:3000/v1
 
 ## enables choosing units and lisenced account call limits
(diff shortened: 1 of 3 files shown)


EgleHelms:
@davehorton @kitajchuk this can be merged. This is just adding a Service Provider ID to when we create carriers regardless via which link. They all still need to be connected to an SP and they were not before, this is why Account user-created ones did not appear in the list of SP providers.

EgleHelms:
To test:
create an account user
login as an account user
Create a Carrier for the account user is responsible for
try to add a Phone number

davehorton (approved):

PR #185, opened 2023-01-24
Spinner spinning forever issue

EgleHelms:
The Spinner issue appeared everywhere where "accounts" array length was needed. So to reproduce this create a new SP and do not create any accounts, then go to Recent Calls, Alerts, Applications where !hasValue(aplications) will be true if accounts array is empty. 

So now this PR is adding a check that also checks for the length of the Accounts. This fixes the issue of the spinner

### DIFF
--- a/src/containers/internal/views/alerts/index.tsx
+++ b/src/containers/internal/views/alerts/index.tsx
@@ -93,7 +93,7 @@ export const Alerts = () => {
       </section>
       <Section {...(hasLength(alerts) && { slim: true })}>
         <div className="list">
-          {!hasValue(alerts) ? (
+          {!hasValue(alerts) && hasLength(accounts) ? (
             <Spinner />
           ) : hasLength(alerts) ? (
             alerts.map((alert) => (
--- a/src/containers/internal/views/applications/index.tsx
+++ b/src/containers/internal/views/applications/index.tsx
@@ -106,7 +106,7 @@ export const Applications = () => {
       </section>
       <Section {...(hasLength(filteredApplications) && { slim: true })}>
         <div className="list">
-          {!hasValue(applications) ? (
+          {!hasValue(applications) && hasLength(accounts) ? (
             <Spinner />
           ) : hasLength(filteredApplications) ? (
             filteredApplications.map((application) => {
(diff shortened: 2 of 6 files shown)


src/containers/internal/views/alerts/index.tsx
```diff
@@ -93,7 +93,7 @@ export const Alerts = () => {
       </section>
       <Section {...(hasLength(alerts) && { slim: true })}>
         <div className="list">
-          {!hasValue(alerts) ? (
+          {!hasValue(alerts) && hasLength(accounts) ? (
```
davehorton, line 96: How is this condition possible, ie to have a non-zero number of alerts but not to have any accounts?  Alerts are associated to an account

EgleHelms: this is possible since the accounts GET call is done first and stored in State, then used in the filter selector to set the state for the selected Account (automatically first in the list). And if there are no accounts that the array will be empty, no account will be selected, thus Alert value will not exist, because the call without an accountSid cannot be made then the Spinner is spinning forever.

So we need to check that the value is non-existent not because of the Account call failing since it doesn't, it is just empty - there are no accounts, thus also no alerts can be fetched.

davehorton: but this whole Section only appears (if I understand) if `hasLength(alerts)`?

EgleHelms: I see what you mean though. the method hasLength is supposed to check if the length is more then 0
`variable.length > minlength`
When in fact empty array is returning true. Hmm, I will look at it more deeply.

EgleHelms: I don't believe so Section component only accepts there the true or false values to apply or not apply certain stylings. Viewing is not affected.

davehorton: ok

davehorton:
@kitajchuk look ok to you?

src/containers/internal/views/alerts/index.tsx
EgleHelms: I also believe there is no value in including both checks either, since as you said Alerts will not exist without accounts and we want to show spinner only when the fetching is in progress. Otherwise, when there are no accounts fetching fails and there is no data.

PR #186, opened 2023-01-25
Combine account user lists for carriers and speech

EgleHelms:
This PR combines Carriers and Speech credential lists for account user and disables Filtering according to the account
Account user views for these lists:

image

image

API server small change to have Speech /SP call act the same way as Carriers: https://github.com/jambonz/jambonz-api-server/pull/98  This way the implementation is the same.

### DIFF
--- a/src/containers/internal/views/applications/edit.tsx
+++ b/src/containers/internal/views/applications/edit.tsx
@@ -34,7 +34,7 @@ export const EditApplication = () => {
 
   return (
     <>
-      <H1 className="h2">Edit Application</H1>
+      <H1 className="h2">Edit application</H1>
       <ApplicationForm application={{ data, refetch, error }} />
     </>
   );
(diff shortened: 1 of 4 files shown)


davehorton:
This is not quite working properly.  It does work properly for carriers, but not for speech.  Here are steps to illustrate:

- as admin user, create an account level user 
- as admin user, add a speech credential at the SP level (in my case, google)
- log out and back in as account user
- create a speech credential (in my case, Azure) and save it
- you are returned to the speech credential list screen with only google showing.  

Note: the Azure credential was in fact added.  If I try to add it again I get a duplicate

davehorton (changes requested):
speech credentials added by account user disappear

davehorton:
Actually, the problem that I had above was related to not running the latest jambonz-api-server.  However, it points up a backwards compatibility issue -- account level credentials added prior to someone upgrading this version will disappear.

EgleHelms:
@davehorton  pointed out a backward compatibility issue this will create. We need to rethink the solution on the webapp and api-server.

davehorton (approved):

PR #188, opened 2023-01-28
feat: add app_json to application

xquanluu:


src/containers/internal/views/applications/form.tsx
```diff
@@ -87,7 +89,7 @@ export const ApplicationForm = ({ application }: ApplicationFormProps) => {
       stateVal: callWebhook,
       stateSet: setCallWebhook,
       initialCheck: initialCallWebhook,
-      required: true,
+      required: callHookRequired,
```
davehorton, line 92: Actually, I think call_hook is still required, even if you put in app_json. The reason is that without this, we won't have a base URL to use to resolve relative URLs in action hooks.  Say for example your app_json has a `gather` which has an `actionHook` property of "/transcript".  Without the base URL from the call hook we would not be able to invoke the webhook in the gather

src/containers/internal/views/applications/form.tsx
```diff
@@ -281,6 +285,20 @@ export const ApplicationForm = ({ application }: ApplicationFormProps) => {
             account={[accountSid, setAccountSid]}
           />
         </fieldset>
+        <fieldset>
```
davehorton, line 288: I'd like a few changes to the UI.  Here is what this looks like now:
image

Please make these changes:
1. Move it to the bottom of the form under the speech credentials
2. Have it hidden by a checkbox labeled "Override webhook for initial application".  This would be like how we do outbound authentication in the Carrier panel for instance:

image

image

3. Make the input a multi-line edit control showing 4 lines.
4. Finally, add an env variable that if set hides this entire feature.  By default it will be available but someone deploying their own system could turn if off globally if they so desired

davehorton (changes requested):

davehorton:
I've opened issue #189 please tag that in future commits on this branch

src/containers/internal/views/applications/form.tsx
xquanluu: fixed

davehorton:
Not quite working yet.  If I create a new application, click the checkbox and add app json and save, it is saved to the database.  However, if I then click on the application to edit it, the checkbox is unchecked so the app json -- which does exist -- is hidden.

Also, the entry field is still an input but I would like it to be a multi-line input

src/containers/internal/views/applications/form.tsx
```diff
@@ -526,12 +531,17 @@ export const ApplicationForm = ({ application }: ApplicationFormProps) => {
...
               label="Override webhook for initial application"
-              initialCheck={applicationJson.length !== 0}
+              initialCheck={initialApplicationJson}
+              handleChecked={(e) => {
+                if (!e.target.checked) {
+                  setApplicationJson("");
+                }
+              }}
             >
               <input
```
davehorton, line 543: This is still an input.  Is it not possible to use a textarea here?

davehorton (changes requested):

davehorton:
@kitajchuk not sure if you have time to take a quick look, but I want to use a textarea in a Checkzone, below is what it looks like if I simply swap the textarea in for the input, would you have any suggestions on improving the UI?

image

davehorton (approved):

davehorton (changes requested):
Currently, if there is an error we get an error message but the input text is removed entirely so we can't easily edit it so as to fix it.  It would be better if when the PUT fails because of an invalid app_json entry that we leave that entry in place so it can be studied and fixed with the help of the error message.

davehorton (approved):

davehorton (approved):

PR #190, opened 2023-01-31
Fix fuzzy search to match against toLowerCase strings

EgleHelms:
In user lists, a fuzzy search was not really accurate when it came to searching by name, since names can have different capitalizations. This PR fixes this issue.

### DIFF
--- a/src/utils/use-filtered-results.ts
+++ b/src/utils/use-filtered-results.ts
@@ -11,7 +11,7 @@ const fuzzyMatch = (patterns: string[], items: string[]) => {
 
   return searchableItems.find((item) => {
     return patterns.some((patternPart) => {
-      return item.toLowerCase().includes(patternPart);
+      return item.toLowerCase().includes(patternPart.toLowerCase());
     });
   });
 };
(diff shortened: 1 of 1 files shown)


src/utils/use-filtered-results.ts
```diff
@@ -11,7 +11,7 @@ const fuzzyMatch = (patterns: string[], items: string[]) => {
 
   return searchableItems.find((item) => {
     return patterns.some((patternPart) => {
-      return item.includes(patternPart);
+      return item.toLowerCase().includes(patternPart);
```
kitajchuk, line 14: We should toLowerCase both sides ya? The values and the pattern input don’t you think?

EgleHelms: @kitajchuk I can add it to be extra careful. But the search bar itself does not allow input of any upper case chars.

davehorton (approved):

src/utils/use-filtered-results.ts
kitajchuk: Oh ya that component does toLowerCase for the controlled input. Still good to add it here so this hook can be used elsewhere if need be so thanks for adding it.

PR #194, opened 2023-02-05
Bug/fix recent calls refresh account assign

EgleHelms:
@davehorton I found a few more issues with the persistent implementation where the accountSid was not always assigned. 

This was mostly related to account user recentCalls/ Alerts views.
page switch when Recent calls is more then the 25 that fit on the page.
Some cleanup in naming.


davehorton (approved):

davehorton (changes requested):
there appear to be conflicts @EgleHelms

davehorton (approved):

PR #201, opened 2023-02-10
feat: nvidia tts/stt speech

xquanluu:


src/vendor/types.ts
```diff
@@ -5,7 +5,8 @@ export type Vendor =
   | "WellSaid"
   | "Nuance"
   | "deepgram"
-  | "IBM";
+  | "IBM"
+  | "NVIDIA";
```
davehorton, line 9: Can you change this to "Nvidia", and while you are at it change "deepgram" to "Deepgram" for consistency

davehorton (changes requested):
Don't we need a list of languages and voices for Nvidia so we can populate the speech dropdowns in the Application form

src/vendor/types.ts
xquanluu: Fixed All review comments

src/vendor/index.tsx
```diff
@@ -39,13 +40,17 @@ export const vendors: VendorOptions[] = [
...
   {
-    name: "deepgram",
+    name: "Deepgram",
     value: VENDOR_DEEPGRAM,
   },
   {
     name: "IBM",
     value: VENDOR_IBM,
   },
+  {
```
davehorton, line 50: The list of vendors is now getting quite long, so whenever we display it in a dropdown please display it in alphabetic order.  This would be on both the speech credentials screen as well as the application screen.

davehorton (changes requested):

davehorton (approved):

Issue #213, opened 2023-03-02
Add support for custom speech vendors


davehorton:
We want customers to be able to add support for abitrary speech vendors by writing to a common API.  As part of this they need to be able to add speech credentials for custom "vendors" in the webapp.  The following changes are needed:

## Adding a speech credential

There should be a new vendor in the dropdown, labeled "Custom".  Since a user should be able to add multiple custom solutions, and we need to distinguish them by a unique vendor name, when the user selects "Custom" we must then gather a "Name" value from them.  When we POST the new credential we will concatenate them as "Custom" colon Name.

For example, the user selects "Custom".  Then next see the entry form with the "Use for text-to-speech" and "Use for speech-to-text" checkboxes.  Directly below that there should be a required input field labeled "Name".  Below that there should be a second (optional) input field labeled "Auth token".  

Finally, if they select to use for STT there should appear an input field labeled  "STT websocket URL" (if STT is selected, this field is required).  And if they selected to use for TTS there should additionally be an input field labeled "TTS http URL".  Again, use for TTS is selected this field is required.

Let's say the user adds a custom speech vendor for both stt and tts, gives a name of 'allomedia' and provides the other information.  The payload should be submitted to the backend as follows:
```json
{
	"vendor": "custom:allomedia",
	"account_sid": "...",
	"service_provider_sid": "...",
	"use_for_tts": 1,
	"use_for_stt": 1,
        "auth_token": "foobarbazzle",
        "custom_stt_url": "wss://myserver.com:2333",
        "custom_tts_url": "http://tts.myserver.com:80"
}  
```
## Editing a speech credential

In the edit panel we should parse the Name from the vendor value returned from the server; e.g. so that we can show "allomedia" in the Name field when the vendor value returned was "custom:allomedia"

davehorton:
hey @xquanluu I tried to rebase your branch on main, where we have added soniox as an stt vendor, but after resolving merge conflicts it looks like I messed it up, because in the Applications view for TTS vendors when I click the pull down I don't see a custom vendor that I added (I do see it in the STT dropdown).  Could you have a look and try to fix what I got wrong, so that this branch is rebased on main (supports soniox STT) and also supports custom vendors ?

xquanluu:
Yes Sure @davehorton, start working on this

xquanluu:
I fixed it @davehorton

xquanluu:
PR merged, close this issue

PR #215, opened 2023-03-03
feat: custom Vendor

xquanluu:


xquanluu:
#213

davehorton:
When I add a custom speech vendor and give it a name ("allomedia"), if I then click to Edit it I do not see the name.  There should be an input field showing "allomedia" just below vendor, otherwise it is unclear what I am looking at.


image

davehorton:
when I select "custom" for TTS for an application, the input fields for Language and Voice correctly change from dropdowns to inputs.  However, the Voice input is pre-filled with a google voice name (en-US-Wavenet-C).  It would be better if it were cleared of any value to start with.  I think it is ok though to leave language defaulting to en-US

image

davehorton:
If I add two custom speech vendors, say "custom:allomedia" and "custom:assemblyai" in the Applications add/edit panel I need to see them both.  Currently I only see just "Custom" and it is unclear which one that is referring to.

And, if I have not added any custom speech vendors I should not see "Custom" in the dropdown

In other words, the dropdown should have one entry for each custom speech vendor added (that supports tts if the tts dropdown, stt if the stt dropdown) with a label like "Custom:allomedia"

src/vendor/index.tsx
```diff
@@ -51,6 +52,10 @@ export const vendors: VendorOptions[] = [
     name: "WellSaid",
     value: VENDOR_WELLSAID,
   },
+  {
```
davehorton, line 55: per the comment in the conversation, it is not that simple.  We can't just add "Custom" to the dropdown because
a) The user may not have added any custom speech providers
b) The user may have added multiple speech providers

We have to build this dropdown dynamically with regard to the custom entries, based on the custom vendors that have been added

davehorton (changes requested):

xquanluu:
> When I add a custom speech vendor and give it a name ("allomedia"), if I then click to Edit it I do not see the name. There should be an input field showing "allomedia" just below vendor, otherwise it is unclear what I am looking at.
> 
> image

Fixed

davehorton:
I'm not sure what that last commit was supposed to fix, but it doesnt seem to have fixed the issues I pointed out with the applications panel, when I go to add or edit I still only have one choice "custom" when I have added 2 custom vendors

image

xanhnguyen:
that's one for application  I'm still fixing @davehorton 
I commited the fix for fixing speech credential page only so that If I do something on application, I still can git hard reset. Let me change the PR to draft PR so don't make notification to you

xquanluu:
that's one for application I'm still fixing [@davehorton](https://github.com/davehorton)
I commited the fix for fixing speech credential page only so that If I do something on application, I still can git hard reset. Let me change the PR to draft PR so don't make notification to you

xquanluu:
Hi @davehorton  I fixed all your comments, Can you help to review once again?

davehorton:
Almost there @xquanluu  but a couple of more changes are needed:

1.  Now when I create 2 custom vendors (allomedia and assembly ai) I *do* see them in the dropdown list in the applications, so that is good.  However, in addition to that I also see a "Custom" entry in the dropdown list -- this needs to be removed.
2. When showing the custom vendors in the dropdown list, I would prefer them to be shown as "allomedia (custom)" to make it clear this is using the custom speech api

xquanluu:
Hi @davehorton I fixed the comments. Could you please help to review again?

davehorton (dismissed):

davehorton:
one last request.  Can you make an env DISABLE_CUSTOM_SPEECH which, if set, hides the "custom" choice when creating speech vendors?

xquanluu:
updating

xquanluu:
@davehorton updated

davehorton (approved):

PR #218, opened 2023-03-24
feat: forgot password

xquanluu:


xquanluu:
Hi @davehorton  could you please also review this PR?

davehorton (changes requested):
We need this feature to be optional -- only shown the user if an env var is configured, because they may not have implemented the mail API interface.

davehorton (dismissed):

davehorton (approved):

PR #223, opened 2023-03-30
Bug: Logout with two clicks

EgleHelms:


davehorton (approved):

davehorton:
Note: this works.  The only slightly off thing is that on logout you get a toast saying "your session expired" which is a bit strange for the user

EgleHelms:
Is this happening when it is deployed? When I run the webapp locally, I am
not getting this.

[image: image.png]

davehorton:
well you already gave me a different PR that fixed this

PR #232, opened 2023-04-08
Feature: Implement Jaeger UI for reading traces

ajukes:
- Added a `View trace` button in recent calls which shows a Jaeger-esq view for reading jaeger traces.
- View trace button does not show if the trace does not exist
- Added a dummy endpoint for dev
- Used Divs and css to build visualization but this may be better using a visualization library however I couldn't find any suitable horizontal waterfall graphs

@xquanluu I am not a react expert or designer and would be great to have some feedback on how I could improve on this first attempt. 

Also visual / layout  logic can probably be improved upon. 
 
Recent calls view:

image

Jaeger Modal:

image

Interactive visualization. Hover and click on spans to view information in details pain.
 
image

Custom horizontal scroll buttons as scrollbar seemed too small:

image


davehorton:
One thing I notice is that in the modal view the start times of all the segments seem to be aligned start-to-start, implying they all started at the same time which is not correct.

ajukes:
I am just cleaning up the logic so will have a look at that. I think they may be different start times but too close to tell.  May need some king of Logarithmic scale or simply extend past the view port for longer durations.

ajukes:
I haven't got it perfect. You can see the offsets more clearly for short duration traces

1m 30 trace:

image

38m trace
image

xquanluu:
Hi @ajukes I'm trying to run your PR locally but seem I cannot see the trace button in every call, Could you please help to fix it then I can easier to review.

ajukes:
Do the ones where you cannot see the trace  have a trace id of 0000000..  ?

If so, there is no trace available to grab.

Otherwise, is an error thrown, or a network failure?

ajukes:
@xquanluu just noticed that calls which are not answered on outdial and a 487 is generated do not have a trace-I'd.

This looks like a bug and may be the reason  you're not seeing the view trace button?

davehorton:
I had a few problems testing this (on jambonz.one, I will send you a login @ajukes).

The main view with the horizontal bars should size the width of the window I feel, not have a horizontal scrollbar -- at least not unless it is smaller than some minimum.  (And I havent tried it out on mobile but it needs to work there too).

As it is, it scrolls way off to the right:
image

Also, when I scrolled over there, I was then unable to scroll back, and got stuck:
image

As far as other UI issues:
- the details panel below the main view is not scrollable, it just extends below the window.  This should actually have a vertical scroll bar I feel
- The line spacing in the details view is too big
- When I drill down to a segment, the "Back" button takes me all the way back to the call details whereas I feel it should bring me back to the toplevel trace view

davehorton:
The changes definitely helped.  A few other things I see:

The back button is clipped:
image

tag values that are booleans (like cached: false) or numbers (like sipStatus:200) are empty in the display but show properly in jaeger.

image

image

image

image

I would still like to see the line spacing a bit tighter if possible.

davehorton:
On mobile view, I think could have the labels view on the left hand side take a bit less space so we get slightly more space for the spans

image

davehorton:
UI is looking good.  Boolean and number tag values still not appearing though

davehorton:
@xquanluu  can you give a review for the code?

davehorton:
Looks like the left hand column of the details pane could use a bit more width
image

ajukes:
What are the window dimensions or phone type that needs the title width increasing?

> Looks like the left hand column of the details pane could use a bit more width image

davehorton:
this was normal desktop view.  You can see the labels in the left column are clipped (e.g. telemetry.sdk.language) while there is plenty of room on the right

image

davehorton:
also the labels and the associated values are not quite horizontally aligned

davehorton:
I feel like there is some bug with calling the `/trace/:trace_id`.

I made a series of calls yesterday.  I am testing with the latest api-server which has your change, and the latest webapp with the last change above.  On call that I can see have traceids, I do not have a view trace button and I can see no `/trace/:trace_id` request is sent from the webapp.

Furthermore, I do not see any pcaps (or requests to `/:call_id/pcap`) **except** when traceid is null.

image

xquanluu:
Hi @davehorton @ajukes  I'm good with the code.

Issue #236, opened 2023-04-13
feature: add support for simple outbound routing in multi-carrier scenario


davehorton:
Currently, when an outbound call is placed if the account has access to multiple outbound carriers at either the SP or account level one is randomly chosen to be used.  We now want to provide support in the portal for a simple version of "least cost routing" (LCR) wherein a service provider or account can create a simple table of digit patterns where an outbound call will select the carrier associated with the first matching digit pattern.

This feature will require [changes to the api-server](https://github.com/jambonz/jambonz-api-server/issues/151) as well (in progress on the feature/lcr branch in that repo).

A service provider user shall be able to create an "LCR table" in the webapp.  This is optional, and if not created then calls will be routed as at present.  When a service provider user creates an LCR table it is associated with the service provider, and it may only reference Carriers created at the service provider level (not account level).

Additionally, an account level user shall be able to create an LCR table.  This is also optional, if an account level user does not create an LCR table then call routing for that account is governed by the service provider LCR table, if any, and by the current random logic if none exists at either level.  When an account level user creates an LCR table it may only refer to Account level Carriers (not service provider level carriers).

>> Note: The reason for the above restrictions is so that each party -- service provider and account -- has control over their costs.

To create an LCR table requires the following steps:

1. Optionally, give a name to the LCR table.
2. Create a default route (/carrier) to be used if there is no digit match (this is required, so we always have a fallback route)
3. Create one or more routes, each route consisting of the following information:

- a digit match, or regex.  If a pattern containing only digits is provided (e.g. 44) then this indicates we will match by seeing if this digit string is a prefix of the called number (e.g., UK in this case).  A regex pattern can also be supplied (no need to provide the bracketing "/"s) and in this case the match is evaluated by testing the regex against the dialed number.
- a carrier; this is the carrier to select if a match is found.

Notes:
1.  The digit matches have an explicit order to them in the LCR table, and the UI should allow the user to drag and drop rows vertically so as to make them be evaluated sooner or later in the process.  The evaluation proceeds from top down as viewed in table.
2. In the initial release, a digit match is associated with a single carrier to use when matched.  In future releases, we may allow more complex scenarios, such as "split traffic 50/50 between these two carriers when the dialed number starts with 44", but this is not needed in the initial release.  (The database schema does support these more complex scenarios though, in expectation of this future work).

PR #237, opened 2023-04-18
feat: Lcr

xquanluu:


xquanluu:
#236

xquanluu:
Hi @davehorton  I have fixed all your comments.

davehorton:
As admin user, I added a lcr table with two routes/digit patterns.  I then used the up arrow to move one of them to the top, saved and came back in.  Now that route has been duplicated and there are 3 routes instead of 2.  In the image below, I entered the pattern `/{10}d/` once, saved and moved it up, saved again, exited and came back and now it has been duplicated

image

davehorton:
The listview looks like this currently
image
To the route of "default account" I would like to show the number of routes, ie "12 routes"

xquanluu:
Screenshot 2023-04-27 at 11 17 27
Hi @davehorton 
I cannot reproduce the issue you mentioned about the dupplication, I have tried multiple times. Can you please try again.

I have edit the list view to have number of routes as requested.

davehorton:
After adding 5 rows, the plus button disappears and I can not add any more.  The requirement is to be able to add essentially an unlimited number of rows.
image

davehorton:
Additionally, I think am not satisfied with the "up" button method of changing the order of the rows.  Could you please see if you can implement drag-and-drop to change the order, instead of using a button.  Perhaps [this react code example](https://react-table-v7.tanstack.com/docs/examples/row-dnd) is useful?

xquanluu:
Start implementing drag and drop inside a scrollable area.

xquanluu:
Hi @davehorton I finished Drag and Drop for LCR route,  unblock unlimited routes, Please help me to review it again.

davehorton:
No, this isn't quite what I had in mind.
image

You can see in the screen shot that *all* rows have a box shadow.  What I would like to see is that *no* rows have box shadow *except* when mouseover a row.  That is, only when mousing over a row should I see it have a box shadow

xanhnguyen:
Screenshot 2023-05-05 at 06 20 53
Hi @davehorton I fixed the shadow box and add transform event that when mouse over, will bring the row up a bit

xquanluu:
Screenshot 2023-05-05 at 06 20 53
I fixed the shadow box and add transform event that when mouse over, will bring the row up a bit

davehorton (approved):

Issue #239, opened 2023-05-01
Admin user can specify domain names for the system


davehorton:
In release 0.8.3 we are adding this table
```
CREATE TABLE system_information
(
domain_name VARCHAR(255),
sip_domain_name VARCHAR(255),
monitoring_domain_name VARCHAR(255)
);
```
The admin user should be able to see and edit these settings in the setting tab.

xquanluu:
Start working on this

PR #240, opened 2023-05-04
#239

xquanluu:
#239


davehorton:
why does the Admin settings show "Service Provider API Keys"?

image

davehorton (approved):

PR #250, opened 2023-05-10
fix: tts.cached boolValue does not showup on spain detail

xquanluu:


davehorton:
OK, in the latest version the mouseover is not working that well.  It is kind of herky-jerky and once the modal dialog is up it seems that mousing off the line does not bring it down.

Let's stick with the modal dialog for now but require a click of the mouse on a tracing line entry to bring up the dialog with the details for that line.  And, once a modal is displayed, a click anywhere outside the modal should then destroy the modal dialog.

Also, the smaller font in the modal dialog looks good, but I would also like to reduce the space between detail lines in the modal.

Finally, I think rather than have the modal dialog expand vertically to fit the content, I think it should have expand only to a fixed size of something like 8 lines of detail and have a scroll bar in the dialog to show more if necessary

xquanluu:
Hi @davehorton 
I have fixed the review comments, could you please help to review again?

davehorton:
Looks much better.  The only problem I see now is that the wrapping of the title on the detail window is not quite right.  In the screen shot below you can see the title extends out past the right hand side of the dialog window, it should wrap.

image

davehorton (approved):

PR #278, opened 2023-06-19
feat: add login with google and github

xquanluu:


davehorton:
@xquanluu I am testing and encountering the following error:

I registered a new account, specifying a sip realm as part of that.  On entry, I select the account screen and make changes to enable call recording.  I click Save, and get this error:

"Use POST /Accounts/:sid/sip_realm/:realm to set or change the sip realm"

I didnt actually modify the sip realm

davehorton:
Downgrade to free plan doesn't work.  I got "something went wrong".  Steps to recreate:

- upgrade to paid plan
- modify payment details, change to new card
- downgrade to trial 

Here are the logs from when I tried to downgrade
```
"msg":"setupFreeTrial: assigning 2c03e0b3-ef6e-425e-9f6c-e3b4834d3f8b to free plan","v":1}

42|jambonz-api-server  | {"level":20, "time": "2023-07-26T12:45:47.198Z","pid":2560775,"hostname":"ip-10-0-214-82","products":[{"product_sid":"2c815913-5c26-4004-b748-183b459329df","name":"registered device","category":"device"},{"product_sid":"35a9fb10-233d-4eb9-aada-78de5814d680","name":"api call","category":"api_rate"},{"product_sid":"c4403cdb-8e75-4b27-9726-7d8315e3216d","name":"concurrent call session","category":"voice_call_session"}],"msg":"setupFreeTrial: added products","v":1}

42|jambonz-api-server  | {"level":20, "time": "2023-07-26T12:45:47.205Z","pid":2560775,"hostname":"ip-10-0-214-82","msg":"setupFreeTrial: deactivated previous plan","v":1}

42|jambonz-api-server  | {"level":20, "time": "2023-07-26T12:45:47.205Z","pid":2560775,"hostname":"ip-10-0-214-82","msg":"setupFreeTrial: deactivating subscription sub_1NY741AxTxXxh2fmOLtGfJkG","v":1}
42|jambonz-api-server  | {"level":30, "time": "2023-07-26T12:45:47.380Z","pid":2560775,"hostname":"ip-10-0-214-82","details":{"error":{"message":"The payment method you provided is not attached to a customer so detachment is impossible.","request_log_url":"https://dashboard.stripe.com/test/logs/req_ldfy0gnMh8X0XU?t=1690375547","type":"invalid_request_error"}},"msg":"downgradeToFreePlan failed","v":1}
```

davehorton:
On the paid plan subscription details, where I enter the number of sessions or registrations if I start by deleting the current value (e.g. the current value is 10 I want to enter 5, I put the cursor in the edit box and click delete twice) I end up with a zero I can not get rid of.  This leads to user interface difficulties.

Per above say I want to change from 10 to 5.  Initially the edit box looks like this:
image

I hit delete twice, to erase the 10, but instead I get a "sticky" 0:
image

Now I enter 5 and it ends up like this:
image


Just get rid of this default zero and let the edit box contents be deleted completely

davehorton:
we also need to get rid of the "null" and replace it with the meaningful unit

davehorton:
I used github oauth register my account.  Therefore I have no password.  But the delete account page requires me to enter a password.  Therefore I can now not delete my account

image

davehorton:
can you fix this so the checkbox is aligned with the associated label
image

davehorton:
If I have a paid plan which includes 5 additional device registrations, then later I want to update my plan to get rid of the 5 additional, I should be able to do that.  Currently, I can not enter "0" in the additional devices entry

image

xquanluu:
Fixed your comments

davehorton:
Clicking on Delete Account does not work -- nothing happens, no api calls made
image

davehorton:
Also, you are only offered the Delete Account option on the Modify Subscription page, which means that people on a free or trial plan can not delete their account.  We should offer them the ability to delete their account as well -- they may want to remove all of their data from the platform

davehorton:
Also, when I upgrade from a trial to a paid account it does work and it does return me to the Account page, but for some reason the view has scrolled down from the top of the account view so I don't immediately see the new account subscription details until I scroll back up to the top, which is slightly confusing because I would like the immediate confirmation that I am now on the paid plan.  The account window I am returned to after upgrading should be scrolled to the top

davehorton (approved):

PR #283, opened 2023-06-30
fix: update name for carrier register authentication fields

xquanluu:
https://github.com/jambonz/sbc-sip-sidecar/issues/35


davehorton:
@xquanluu can you rebase?

PR #329, opened 2023-10-12
upgrade wavesuffer to 7.3.4

xquanluu:


xquanluu:
Screenshot 2023-10-12 at 12 41 14

src/containers/internal/views/recent-calls/player.tsx
```diff
@@ -46,15 +46,19 @@ export const Player = ({ call }: PlayerProps) => {
 
   const wavesurferId = `wavesurfer--${call_sid}`;
   const wavesurferTimelineId = `timeline-${wavesurferId}`;
-  const waveSufferRef = useRef<WaveSurfer | null>(null);
+  const waveSuferRef = useRef<WaveSurfer | null>(null);
```
davehorton, line 49: can you fix the typos?  "waveSufer" should be "waveSurfer"

davehorton (changes requested):

davehorton:
There is no change in functionality here, is that right?  Just upgrading to the most recent version of wavesurfer?

xquanluu:
yes it's correct

The timeline ruler is more detail in 0.2 second

davehorton:
ok great.  After fixing the comment above can you deploy to jambonz.me for testing?

xquanluu:
fixed typo issue

davehorton (approved):

Issue #328, opened 2023-10-09
Add ElevenLabs as TTS vendor


davehorton:
https://elevenlabs.io/speech-synthesis

davehorton:
Eleven Labs list of voices [here](https://gist.github.com/davehorton/05618fda0ab360f91240da3d2edf0ed5).

ElevenLabs does not have a separate list of voices per language.  Instead, each language they support has the same list of voices.  Getting the list of languages they support is slightly different than other synthesizers.  They support a few "models" and each model has a list of languages that is supported.  THe current models are:
- eleven_multilingual_v2
- eleven_multilingual_v1
- eleven_monolingual_v1 (English only)

To start with we should offer all of the languages supported by the eleven_multilingual_v2 model.  Details on the models can be found [here](https://gist.github.com/davehorton/4df0a704624a770c76ea212da4fb2a06), including the list of languages.

So for the Applications page, when someone selects ElevenLabs for speech synthesis, show all 29 languages that are supported by the eleven_multilingual_v2 model.  The language dropdown should show the language name (e.g. English) and store the associated code as defined in the model as language_id (e.g. en).

Regardless of the language selected, the same list of voices should be shown.  Per the list of voices in the link above, each voice has a name (Rachel) and a voice_id (21m00Tcm4TlvDq8ikWAM).  The voice id is what should be stored in the database.

davehorton:
For testing create a free account https://elevenlabs.io/sign-up

PR #330, opened 2023-10-12
add elevenlabs

xquanluu:
https://github.com/jambonz/jambonz-webapp/issues/328


davehorton:
Change Model Id to just Model, and show the other two models as well
image

davehorton:
The names should just be "multilingual v2", "multilingual v1", and "monolingual v1"

xquanluu:
fixed

davehorton:
In the dropdown list of voices in the Application view when you select ElevenLabs you currently get a list of voice names, ordered as they are returned from ElevenLabs in the GET /voices.  Instead, please show them in alphabetic order.  Also, instead of 
```
Rachel
```
show it as
```
Rachel - american, calm, young, female, narration
```
where those values after the dash are the values in the `labels` property of the voice object that is returned, e.g.
image

davehorton:
If the user chooses "monolingual v1" as the model, only the English language should be offered

src/vendor/speech-synthesis/elevellabs-speech-synthesis-lang.ts
```diff
@@ -0,0 +1,191 @@
+import type { VoiceLanguage } from "../types";
+
+export const languages: VoiceLanguage[] = [
```
davehorton, line 3: Since ElevenLabs has a public api to get voices, instead of hardcoding this list, can we retrieve it dynamically when the Application view comes up - or probably better yet when the speech synthesis dropdown on the Applications panel is selected to be Eleven Labs?

The GET /Voices api can be done with no authentication - in which you get back the standard list of voices -- or it can be done with the api-key of the user, in which case you get back the standard voices plus any cloned voices that the user has added.  So we want to use the authenticated version, grabbing the api key that the user has entered (this presumes that if they select ElevenLabs on the Application view they have previously entered a speech credential for Eleven Labs -- if this is not the case, then just use the unauthenticated version of the API and go with that.

So the api call (in curl) will look like this
```
curl -X 'GET' \
  'https://api.elevenlabs.io/v1/voices' \
  -H 'accept: application/json' \
  -H 'xi-api-key: <api-key>'
```

I will give you my api key separately to use for testing, as I have cloned my own voice and added it to my account

davehorton (changes requested):

davehorton:
Also, one further request.  In the Voices payload response there is a property called "category".  When creating the display name in the dropdown, if the category value is anything other than "premade", please put it next to the name in parentheses.

For instance, Rachel is a premade name so it would look like as I suggested above:
```
Rachel - american, calm, young, female, narration
```
However, DaveH is a cloned voice:
```
        {
            "voice_id": "TwVPyTHwMlVlXTKVW83m",
            "name": "DaveH",
            "samples": [
                {
                    "sample_id": "klBmsEttYjvdik5UYK9C",
                    "file_name": "dave-3mins.mp3",
                    "mime_type": "audio/mpeg",
                    "size_bytes": 646848,
                    "hash": "a50c833c62f988b25b28f42ce4bad0a2"
                }
            ],
            "category": "cloned",
            "fine_tuning": {
                "language": null,
                "is_allowed_to_fine_tune": false,
                "fine_tuning_requested": false,
                "finetuning_state": "not_started",
                "verification_attempts": null,
                "verification_failures": [],
                "verification_attempts_count": 0,
                "slice_ids": null,
                "manual_verification": null,
                "manual_verification_requested": false
            },
            "labels": {
                "accent": "American",
                "sex": "male",
                "type": "slightly raspy",
                "age": "50+"
            },
```
So it should appear as:
```
DaveH (cloned) - American, male, slightly raspy, 50+
```

davehorton (dismissed):

davehorton:
@xquanluu there is a bug when setting a voice in the Application view.  When I save it, it does use that name but when I come back in again to the Application page it shows the voice as the first one in the list (Adam) not the one I selected

https://www.loom.com/share/864a6003e49b4e3cb36dabf6287880f3?sid=3a6d9145-285a-42da-979f-565294bd3707

xquanluu:
Fixed

davehorton (approved):

PR #341, opened 2023-11-03
feat calculate speech recognition latency

xquanluu:


src/containers/internal/views/recent-calls/player.tsx
```diff
@@ -135,7 +135,7 @@ export const Player = ({ call }: PlayerProps) => {
               count++;
               if (count === PEAKS_WINDOW) {
                 return (
-                  ((i + PEAKS_WINDOW) * duration) / data.length + 20 / 1000 // this is 20 ms adjustment
+                  ((i + PEAKS_WINDOW) * duration) / data.length + 40 / 1000 // this is 20 ms adjustment
```
davehorton, line 138: please make 50

davehorton (changes requested):

src/containers/internal/views/recent-calls/player.tsx
xquanluu: updated

PR #357, opened 2023-11-30
elevenlabs new model and options

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/541

### DIFF
--- a/src/api/constants.ts
+++ b/src/api/constants.ts
@@ -1,5 +1,6 @@
 import type {
   Currency,
+  ElevenLabsOptions,
   LimitField,
   LimitUnitOption,
   PasswordSettings,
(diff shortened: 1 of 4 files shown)


davehorton:
- change wording to "Beautify"
- can we add an anchor tag with label "Docs" that goes to the url https://elevenlabs.io/docs/api-reference/streaming ?  I think if we align it horizontally with "Extra options" but push it to the right so that it above "Beautify" that would be good

xquanluu:
Screenshot 2023-11-30 at 22 09 08

xquanluu:
Extra Options is header of CheckZone, modify it is quite hard, so I put Docs and Beauty at the same row

davehorton:
@xquanluu can you deploy to jambonz.me for testing?  In the screen shot "Docs" and "Beautify" seem like they are not quite aligned horizontally

davehorton (approved):

Issue #358, opened 2023-12-01
Show latency for TTS on the waveform


davehorton:
Similar to how we show the STT latency, we should also show the latency for TTS.  Hopefully this will not be too difficult since we already have the tts latency reported on the tracing spans.

The latency for TTS should be displayed similarly (in the form of a region) but it should be on the lower audio track since that is where the audio playback is occurring.  The region should be clickable and when you click it you should get a popup showing:
- tts vendor
- latency
- whether it was retrieved from cache

Also, we should one color for the regions for TTS which was generated, and a different color for TTS which was retrieved from cache.  As with the STT latency we should display the latency in secs next to each region.

xquanluu:
![Screenshot 2023-12-04 at 23 45 08](https://github.com/jambonz/jambonz-webapp/assets/110280845/2dda48f9-5e47-40f5-bbe6-e5f293b89be8)
Deployed to jambonz.me

PR #359, opened 2023-12-04
tts latency feature

xquanluu:
#358


davehorton:
actually, I think we should **only** show TTS latency when we have actually generated it from the TTS service. If it was retrieved from cache, dont show a region at all

xquanluu:
I fixed review comments, deployed to jambonz.me

davehorton (approved):

Issue #362, opened 2023-12-12
Add support for outbound OPTIONS ping from carrier trunk


davehorton:
We should have an option for outbound gateways to send OPTIONS pings to the remote gateway.  This of course is only valid for sip gateways that are outbound.  There should be a checkbox for each outbound gateway in the webapp labeled "send OPTIONS ping".  If checked, the system (sbc-sip-sidecar) should send an OPTIONS ping every 60 seconds (this should be an env var) to the gateway.  If the gateway fails to respond, or responds with a non-success response, the corresponding sip gateway should be set to inactive (we can leverage the is_active column that already exists in the database).  When a gateway transitions from active to inactive due to a failure of an OPTIONS ping we should write an alert

Megamuch:
I would like to add that a single failure might be too strict for disabling the gateway. I would keep a minimum of 5 missed pings so you really know it is down. 

Another thing to think about is recovery. What happens is the gateway goes down for 1 hour and comes back up. Do we require a manual intervention into jambonz to re-activate the gateway?

vdharashive:
@xquanluu and @davehorton can this feature be completed , its useful feature to keep the trunk up

davehorton:
yes, this requires database changes so we deferred it to 0.8.6 which we will start shortly

PR #363, opened 2023-12-12
Send options ping sip gateway

xquanluu:
https://github.com/jambonz/jambonz-webapp/issues/362

Screenshot 2023-12-12 at 14 02 48


src/containers/internal/views/carriers/form.tsx
```diff
@@ -1121,6 +1121,29 @@ export const CarrierForm = ({
                           </label>
                         </div>
                       )}
+                      {Boolean(g.outbound) && (
+                        <div>
+                          <label
```
davehorton, line 1126: I dont know where the label text is kept but I would like it to be "send OPTIONS ping"

davehorton (changes requested):

src/containers/internal/views/carriers/form.tsx
xquanluu: the label text is here: https://github.com/jambonz/jambonz-webapp/pull/363/files#diff-512d63808fe31447ace3dfaffff744446e16489c2f767085211a9d70d2f300deR1141

And Yes this just apply to outbound gateway only

davehorton (approved):

PR #373, opened 2023-12-28
Update with all supported google TTS languages

vdharashive:


davehorton (changes requested):
please use the same settings as we do for tabs (tabs to spaces, indent 2) so we get a proper diff

Issue #378, opened 2024-01-11
Speech label selector cases that dont work


EgleHelms:
**Case 1**
Repro steps:

1. create MS cred with no label and choose "I want to use a custom voice for TTS" and add a deployment id
2. create MS cred with "aa" label (no custom voices)
3. go to application page

Result: preselected label will be "None" but the "Voice" will not appear as a text field for custom voice

Also, I think voice field should only show a placeholder, it should not have a voice prefilled, because this is misleading. From what I have tested so far, the MS credentials that will hold custom voice will not have other voices available, so this field should not be prefilled, just have a placeholder

**Case 2**
1. create credential with label1
2. create a credential with label2
3. select credential label1 in the Application form and save
4. Delete credential label1 (in MS credential case change to have custom)
5. go back to Application form see label2 preselected and save

Result: the label1 will still be as label in the api-server 

Case 3: 
it is currently possible to have credentials without labels in the same level (account or all accounts) as well as on both. No way it is possible to know what will be applied.
I think it should not be allowed to have no label 2 times regardless of level.

=======
In addition we need to make sure none of the other cases are broken:

Standard Cases:
1. having one credential with no label
2. having 2 credentilas with no label should not be possible
3. having 2 credentials with labels
4. having 2 credentials one with label, one without
5. what appears preselected is also savable
6. having an application that already had a credential setup, then the preselected is savable
7. Important to check if payload sent and what UI shows matches
8. if MS creds have custom voice the "Voice" field should always be rendered

Could also be an improvement to list the labels, speech providers alphabetically.

davehorton:
@xquanluu can you try to recreate?

xquanluu:
Hi @davehorton 

case 1: Could you send me in private Message for an credentail with deployment ID.
case 2: fixed, PR is in draft version
case 3: This is exactly the same the case we add one service provider level google credential without label and add another account level google credential without label. API server and feature server will choose account level first, if there is no account level credential, it will choose service provider level.

xquanluu:
Hi @davehorton @EgleHelms 
The PR fixes Case 1 and Case 2.

For Case 3 Could you please check my previous comment?

davehorton:
for now I thinkk case 3 is a suggested enhancement, not a bug so I am merging this

PR #384, opened 2024-01-16
fix remove speech label, cause application save wrong label

xquanluu:
https://github.com/jambonz/jambonz-webapp/issues/378


xquanluu:
Fixed Case 2

xquanluu:
Fixed Case 1

davehorton:
please rebase and set for review if cases 1 and 2 in the issue have been tested

Issue #397, opened 2024-02-06
add support for deepgram on prem


davehorton:
Similar to Microsoft, we need an option for hosted deployment.  When hosted deployment is selected we should collect two pieces of information:
- deepgram_uri - a string in the form "ip:port"
- use_tls - a checkbox that defaults to off

Of course, this must be backwards compatible - ie existing deepgram credentials should continue working against their hosted system

PR #398, opened 2024-02-07
Support deepgram onprem

xquanluu:
https://github.com/jambonz/jambonz-webapp/issues/397

Screenshot 2024-02-07 at 19 00 54


davehorton (dismissed):

davehorton:
Please change "Use hosted Deepgram service" to "Use on-prem Deepgram container"

davehorton:
@xquanluu api key is *not* required when on-prem is used.  It is optional in that case

Issue #417, opened 2024-04-04
Add support for registering over SIP/TLS


davehorton:
When creating a carrier that requires registration, we need to allow the user to specify that the REGISTER requests should be sent over SIP/TLS.  A checkbox, defaulting to off, should do it.  This will require changes to the database schema and api server as well.

PR #418, opened 2024-04-04
register use tls

xquanluu:
#417 

Screenshot 2024-04-04 at 15 22 53


davehorton:
Please change label from "Use TLS" to "Use SIP over TLS"

davehorton (approved):

davehorton (approved):

PR #440, opened 2024-06-15
Add support for both legacy jaeger and latest otel trace formats

zscgeek:
Added support to the front end so it supports the legacy Jaeger format and the new otel format returned by the latest versions of Jaeger.


zscgeek:
I rebased on the latest changes to make this cleaner to merge.

davehorton:
can you provide some details on what is necessary to configure a jambonz system to use otel format, in terms of env vars for feature-server container and other components that need to be running @zscgeek

Issue #456, opened 2024-10-04
Allow account to set a log level for application logs


davehorton:
A great suggestion was made by a customer recently.  For troubleshooting purposes, it would be great to allow someone in the portal to set the loglevel for the account.  This could be a checkbox to enable debug logging.  This log level would apply specifically to the jambonz-feature-server, sbc-inbound, and sbc-outbound since those are the logs we use most often in troubleshooting call issues.  Those corresponding apps would retrieve the log level as part of the current account retrieval and then set the pino loglevel accordingly, just for that session.

That way we can get debug logs for one customer without having to turn up for all, and also without having to restart the apps.

As part of this, in system setttings we should probably have a setting for default loglevel - dropdown of info or debug, to set/change the system wide settings.

xquanluu:
Tested them all on jambonz.me

PR #457, opened 2024-10-07
allow system information contains log level and account has enable_debug_log

xquanluu:
https://github.com/jambonz/jambonz-webapp/issues/456
Screenshot 2024-10-07 at 16 00 40
Screenshot 2024-10-07 at 16 00 30


davehorton:
can we hide the account-level setting unless the user had admin priviledges?

xquanluu:
fixed, only >= Admin can set account log level

davehorton (approved):

PR #458, opened 2024-10-10
support speechmatics speechcredential

xquanluu:


davehorton:
sorry @xquanluu we also need to store the [url of the endpoint](https://docs.speechmatics.com/introduction/authentication#supported-endpoints) as it turns out.

On that page you can see there are currently 3 supported endpoints for their streaming api.  
- EU: eu2.rt.speechmatics.com (on demand)
- EU: neu.rt.speechmatics.com (enterprise customers only)
- US: wus.rt.speechmatics.com

Let's have a region dropdown with those 3 choices, showing in the same way, e.g.

"EU: eu2.rt.speechmatics.com (on demand)"

as far as what we send to the api-server, send the URL only.  

Also please put the 3 choices in some sort of data file or structure that will be easy to update in the future, as I am sure they will add new endpoints

davehorton:
actually, I notice they support on-prem as well.  So I think we need
- radio buttons for hosted or on-prem
- when hosted selected, dropdown with the 3 URLs
- when on-prem selected, simple input field for URL

plus api key regardless of above

xquanluu:
Screenshot 2024-10-11 at 11 08 11
Screenshot 2024-10-11 at 11 03 32

davehorton:
Looks good @xquanluu.  One last request: could you add a variable to the .env called VITE_APP_DISABLE_ADDITIONAL_SPEECH_VENDORs and if set to true do **not** include speechmatics in the dropdown list.

davehorton (approved):

PR #476, opened 2025-01-09
support custom tts streaming

xquanluu:
https://github.com/jambonz/jambonz-feature-server/pull/1023

Screenshot 2025-01-09 at 13 00 23

### DIFF
--- a/src/containers/internal/views/speech-services/form.tsx
+++ b/src/containers/internal/views/speech-services/form.tsx
@@ -939,7 +939,7 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
                   }}
                 >
                   <label htmlFor="custom_vendor_use_for_tts">
-                    Http URL (non-streaming)<span>*</span>
+                    Http URL (non-streaming)
                   </label>
                   <input
                     id="custom_vendor_use_for_tts"
@@ -953,7 +953,7 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
                     }}
                   />
                   <label htmlFor="custom_vendor_use_for_tts_streaming_ws">
-                    Ws URL (streaming)<span>*</span>
+                    Ws URL (streaming)
                   </label>
                   <input
                     id="custom_vendor_use_for_tts_streaming_ws"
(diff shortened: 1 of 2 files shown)


xquanluu:
Hi @davehorton 

I have updated the UI as your comments
Screenshot 2025-01-14 at 14 52 15

Http url or ws url, one of them will be required if the other is missing.

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -914,26 +925,47 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
...
                       setCustomVendorTtsUrl(tmpCustomVendorTtsUrl);
+                      setCustomVendorTtsStreamingUrl(
+                        tmpCustomVendorTtsStreamingUrl,
+                      );
                     }
                   }}
                 >
                   <label htmlFor="custom_vendor_use_for_tts">
-                    TTS HTTP URL<span>*</span>
+                    Http URL (non-streaming)<span>*</span>
```
davehorton, line 942: hmm, I think we should now remove the asterisk, since it is possible to only add the other?

src/containers/internal/views/speech-services/form.tsx
```diff
@@ -914,26 +925,47 @@ export const SpeechServiceForm = ({ credential }: SpeechServiceFormProps) => {
...
                     placeholder="Required"
-                    required={ttsCheck}
+                    required={ttsCheck && !customVendorTtsStreamingUrl}
                     value={customVendorTtsUrl}
                     onChange={(e) => {
                       setCustomVendorTtsUrl(e.target.value);
                     }}
                   />
+                  <label htmlFor="custom_vendor_use_for_tts_streaming_ws">
+                    Ws URL (streaming)<span>*</span>
```
davehorton, line 956: ditto

davehorton (changes requested):

src/containers/internal/views/speech-services/form.tsx
xquanluu: fixed

src/containers/internal/views/speech-services/form.tsx
xquanluu: fixed

davehorton (approved):

Issue #486, opened 2025-02-25
Add logs panel to Recent Calls


davehorton:
Using the REST api defined in [this issue](https://github.com/jambonz/jambonz-api-server/issues/392),  the Recent Calls detail panel should have new tab to the right of "Details" and "Tracing" called "Logs".  This tab should only be enabled if the env var `AWS_ENABLE_CLOUDWATCH_LOGS` is defined for the webapp.

When you click the Logs tab, there should be a button "Retrieve Logs" and an initially empty content window.  If the user clicks the button and we get a 403 error back we should inform the user that logs are not available on this system; if no logs are returned (empty output) we should inform the user that no logs were found.  If logs are returned we should show them in the content window which should also have a button to copy them to clipboard.

We should also have text or information on the panel somewhere informing the user that only logs from within the past 3 hours can be retrieved.

PR #490, opened 2025-02-27
support showing call cloudwatch logs

xquanluu:
#486 

Screenshot 2025-02-27 at 17 40 49


davehorton:
A few things.  

First, when looking at the scrollable logs window, I would like there to be small button icons just above the scrollable text for "copy" and "download".  A good example of what I am looking for is the Claude.ai interface that it shows when it generates a source code file.

image

Second, we can then get rid of the Download Logs button

Third, when the logs feature is not available (i.e on a non-AWS system) instead of showing "Logs" show in a greyed out/ disabled state "Logs (AWS only)"

Fourth, when the logs are downloaded we lose the formatting.  Instead, when downloading the logs, or when copying them to the clipboard please first run them through the equivalent of `pino-pretty -c -t`

davehorton:
Also I don't love the bright green text color. Lets try to match the colors in that Claude snippet above -- background is a grayish-black, text is the whitish color of "Async3CompatibleUploader"

xquanluu:
Fixed review comments.

I cannot run pino-pretty lib on web browser

Screenshot 2025-02-28 at 10 21 22

davehorton:
The logs now have timestamps, but I would like the timezone to be GMT regardless of where the user is
```
{
  "level": 30,
  "time": "2025-03-04 16:06:02-05:00",
```

davehorton:
what does it mean when "Retrieve Logs" is grayed out
image

xquanluu:
> The logs now have timestamps, but I would like the timezone to be GMT regardless of where the user is
> 
> ```
> {
>   "level": 30,
>   "time": "2025-03-04 16:06:02-05:00",
> ```

fixed

Screenshot 2025-03-06 at 11 47 45
Screenshot 2025-03-06 at 11 51 30

xquanluu:
Hi @davehorton 

I have deployed this to jambonz.one. please let me know if you need a change for this

davehorton (approved):

davehorton:
rebase

PR #493, opened 2025-03-17
fix: application synthsizer configuraiton is shown different than data stored in db  when view application.

xquanluu:
Step to reproduce:
1/ create cartesia speech credntial
2/ update application to use cartesia
3/ application will choose default english language and voice
4/ change the language to french and french voice as well.
5/ save the application and go to Recent call
6/ go back to recent call and see that the synthesizer is change to default english again but data return from backend still showing fr language and voice.


xquanluu:
Tested on jambonz.xyz and it's working

davehorton (approved):

Issue #495, opened 2025-03-23
add support for openai as stt vendor


davehorton:
Add support for new STT vendor OpenAI, see [here](https://platform.openai.com/docs/guides/speech-to-text) for details.

When configuring the speech vendor
- user should provide api key
- user should select from one of the 3 models: whisper-1, gpt-4o-mini-transcribe, gpt-4o-transcribe
- there is no on-prem option
- we should test with a static wave file bundled with api server

xquanluu:
closing this as PR is merged

PR #496, opened 2025-03-24
support openAi stt

xquanluu:
#495 

Screenshot 2025-03-24 at 16 23 04


davehorton:
we should not require the user to specify the model as part of the credential.

xquanluu:
Fixed, deployed to jambonz.xyz

davehorton:
@xquanluu I'm sorry, I changed my mind on saving the model as part of the credential.  In looking at the code and other vendor examples, I now do think we need to save that.  Could you add that back in?

xquanluu:
added openai model selection. deployed to jambonz.xyz

davehorton (approved):

PR #500, opened 2025-04-01
fixed error when selecting yesterday option from date filter

rammohan-y:
fixed an error when "yesterday" option is selected as the api is called with NaN. Also added all options from the DATE_SELECTION constants

@xquanluu @davehorton @vinoddharashive


xquanluu:
Look good to me.

davehorton (approved):

PR #503, opened 2025-04-11
Disable password managers on some password forms

sammachin:
add the data-lpignore attribute to the password component and remove redundant {} on autoComplete.

This should resolve the recurring issue we've had with users having issues creating a carrier where the page was expecting an SMPP IP adddress, that issue seemed to be that lastpass was adding their jambonz password as an SMPP password so the form validation expected an SMPP IP to go with the credentials.

By default lastpass don't honor the autocompete=off attribute unless you change your settings but they also have this aditional attribute which should then ignore the field for all lastpass users 
https://support.lastpass.com/s/document-item?language=en_US&bundleId=lastpass&topicId=LastPass/c_lp_prevent_fields_from_being_filled_automatically.html&_LANG=enus

### DIFF
--- a/src/components/forms/passwd/index.tsx
+++ b/src/components/forms/passwd/index.tsx
@@ -32,7 +32,7 @@ export const Passwd = forwardRef<PasswdRef, PasswdProps>(
       <div className="passwd">
         <input
           autoComplete="off"
-          data-lpignore=true
+          data-lpignore="true"
           ref={ref}
           type={reveal ? "text" : "password"}
           name={name}
--- a/src/components/forms/passwd/index.tsx
+++ b/src/components/forms/passwd/index.tsx
@@ -31,7 +31,8 @@ export const Passwd = forwardRef<PasswdRef, PasswdProps>(
     return (
       <div className="passwd">
         <input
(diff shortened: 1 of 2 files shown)


davehorton:
let us know when this has been tested

sammachin:
sorry thought it was still in draft, yes testing now

sammachin:
Added config for the other common password managers (1password, bitwarden & dashlane) we know LastPass was causing issues but good to set them all I think. Info from https://www.stefanjudis.com/snippets/turn-off-password-managers/

I've defaulted to disabling password manager on passwd components as we use it internally in several places like carrier settings, then I've explicitly re-enabled it on the login form,

sammachin:
@davehorton tested and ready for review

davehorton:
thanks sam.  And these new attributes ("data-lpignore" etc) work on all browsers and wont create problems if they are unsupported?

sammachin:
yes they're the reccomended attributes for site owners to set by the various password manager platforms, no issues when I tested on Chrome & Firefox so thats the 2 key rendering engines covered

davehorton (approved):

Issue #504, opened 2025-04-25
add support for newest Rime labs model - arcana


davehorton:
arcana api details
https://docs.rime.ai/api-reference/arcana/streaming-mp3

voice details
https://docs.rime.ai/api-reference/voices
https://users.rime.ai/data/voices/all-v2.json
https://docs.rime.ai/api-reference/data/voices
https://users.rime.ai/data/voices/voice_details.json
https://docs.rime.ai/api-reference/data/voice-details

xquanluu:
Hi @davehorton 
Only https://gitlab.com/jambonz/freeswitch-modules/-/merge_requests/86 is required for review

PR #510, opened 2025-05-06
support rimelabs arcana

xquanluu:
https://github.com/jambonz/jambonz-webapp/issues/504

### DIFF
--- a/src/containers/internal/views/applications/speech-selection.tsx
+++ b/src/containers/internal/views/applications/speech-selection.tsx
@@ -290,7 +290,11 @@ export const SpeechProviderSelection = ({
             return;
           }
           if (synthVendor === VENDOR_RIMELABS) {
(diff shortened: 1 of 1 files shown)


src/containers/internal/views/applications/speech-selection.tsx
```diff
@@ -290,7 +290,11 @@ export const SpeechProviderSelection = ({
             return;
           }
           if (synthVendor === VENDOR_RIMELABS) {
-            const newLang = json.tts.find((lang) => lang.value === "eng");
+            let newLang = json.tts.find((lang) => lang.value === "eng");
```
davehorton, line 293: this looks like some sort of workaround?

xquanluu: No it's not a workaround. Arcana model support **Any** language, so there is only one language any for all the available voices.

Issue #523, opened 2025-05-22
some S3 compatible storage systems have a region parameter


davehorton:
Currently we only allow region to be entered when AWS S3 is used.  However, some - but not all - S3 compatibles also have a region that we need to use when making requests.  So we when S3 compatible is selected we should show an input field for region -- not a dropdown.  It is optional but the user can enter a region if their storage requires it.

PR #524, opened 2025-05-23
some S3 compatible storage systems have a region parameter

xquanluu:
#523


xquanluu:
@davehorton 
this PR include a proper way to fix toastError and toastSuccess.

currently, in main branch, the webapp is using global react state to store toast message [code here](https://github.com/jambonz/jambonz-webapp/blob/main/src/store/index.tsx#L93) developed inside StateProvider component. when toast message is set to this global react state, it make the current page re-render again and clear all of user input values for that page.

With new way, I introduced new toastProvider which has it own react context, that allow the toast running separately with the global react state and allow the app to show toast message without re-rendering the page again and now webapp does not reset user input value when toast message is shown

PR #551, opened 2025-07-20
Draw STT latency to recording player by using stt metrics from opentelemetry

xquanluu:
Screenshot 2025-07-20 at 10 21 49

Screenshot 2025-07-20 at 10 21 25


davehorton:
does this fall back to using the older method if span data does not exist?

xquanluu:
Hi @davehorton 
Yes It will fallback to the legacy method if there is no stt.latency_ms available in the metric

PR #577, opened 2025-11-10
fix outbound call routing race condition show default lcr route set that user cannot delete

xquanluu:
https://github.com/jambonz/jambonz-api-server/issues/510


davehorton:
please rebase

